PR #78, opened 2016-04-22
Added billing info to client model

### LINKED ISSUE #29: Add type of facturation on member
When a client is add on the system, he may pay different prices.

## Task

 - Add CharField-Choice on client

## Choices

 - Defaults
 - Low income
 - Solidary

adasilva:
fixed #29


django/santropolFeast/member/models.py
```diff
@@ -159,12 +159,25 @@ class Client(models.Model):
...
+        ('solidary', _('solidary')),
+    )
+
     class Meta:
         verbose_name_plural = _('clients')
 
     billing_address = models.ForeignKey(
         'member.Address',
-        verbose_name=_('billing_Address')
+        verbose_name=_('Billing Address')
```
RignonNoel: Be carefull we do not write in English but we just create id which will then be translated.
Here ```billing_address``` is the good code

django/santropolFeast/member/models.py
```diff
@@ -159,12 +159,25 @@ class Client(models.Model):
         (DECEASED, _('deceased')),
     )
 
+    FACTURATION_TYPE = (
+        ('default', _('default')),
+        ('low income', _('low income')),
```
RignonNoel: Be carefull we do not write in English but we just create id which will then be translated.
Here you need to have ```low_income``` per example

django/santropolFeast/member/models.py
```diff
@@ -159,12 +159,25 @@ class Client(models.Model):
...
         verbose_name_plural = _('clients')
 
     billing_address = models.ForeignKey(
         'member.Address',
-        verbose_name=_('billing_Address')
+        verbose_name=_('Billing Address')
+    )
+
+    facturation = models.CharField(
+        verbose_name=_('Billing type'),
```
RignonNoel: Be carefull we do not write in English but we just create id which will then be translated.

django/santropolFeast/member/models.py
```diff
@@ -174,7 +187,7 @@ class Meta:
 
     emergency_contact = models.ForeignKey(
         'member.Member',
-        verbose_name=_('emergency_contact'),
+        verbose_name=_('Emergency Contact'),
```
RignonNoel: Be carefull we do not write in English but we just create id which will then be translated.

RignonNoel:
Thank you very much @adasilva ! It's very sympathic to help us in the project.

 1- I've just annoted your code, it need little correction.
 2- To pass test you just need to add migrations file in a new commit.

```
cd /path/to/the/project/directory
python manage.py makemigrations
```

-------

The rest is ok for me but I would just raise some points of good practice that are obligatory in more serious projects:

 - You can't make a commit just to correct PEP8, but you can amend a commit ```git commit --amend```
 - You can't modify the login page in this PR because is not a task of the issue. It was harder to track changes and emerging errors may occur.

PR #89, opened 2016-04-26
fix 88 gender name

### LINKED ISSUE #88: Fix the gender name in the member model
The name are presently in french they need to be in english.

maximest-pierre:
fixed #88 and also change the default value to unknown.

### DIFF
--- a/django/santropolFeast/member/models.py
+++ b/django/santropolFeast/member/models.py
@@ -36,11 +36,11 @@ class Member(models.Model):
     gender = models.CharField(
         max_length=1,
         choices=(
-            ('M', 'Male'),
-            ('F', 'Female'),
-            ('U', 'Unknown'),
+            ('M', _('male')),
+            ('F', _('female')),
+            ('U', _('unknown')),
         ),
-        default=2
+        default='U'
     )
 
     birthdate = models.DateField(
--- a/django/santropolFeast/member/models.py
+++ b/django/santropolFeast/member/models.py
@@ -36,11 +36,11 @@ class Member(models.Model):
     gender = models.CharField(
         max_length=1,
         choices=(
-            ('H', 'Homme'),
-            ('F', 'Femme'),
-            ('U', 'Inconnu'),
+            ('M', 'Male'),
+            ('F', 'Female'),
+            ('U', 'Unknown'),
         ),
-        default=1
+        default=2
     )
 
     birthdate = models.DateField(


django/santropolFeast/member/models.py
```diff
@@ -36,11 +36,11 @@ class Meta:
...
         choices=(
-            ('H', 'Homme'),
-            ('F', 'Femme'),
-            ('U', 'Inconnu'),
+            ('M', 'Male'),
+            ('F', 'Female'),
+            ('U', 'Unknown'),
         ),
-        default=1
+        default=2
```
RignonNoel: I'm not sure on this default value:

 - When i create a member in the python shell, the member.gender value is u'2'
 - When i create a member in the admin panel, the default value in the dropdown is 'Male'

django/santropolFeast/member/models.py
```diff
@@ -36,11 +36,11 @@ class Meta:
     gender = models.CharField(
         max_length=1,
         choices=(
-            ('H', 'Homme'),
-            ('F', 'Femme'),
-            ('U', 'Inconnu'),
+            ('M', 'Male'),
```
RignonNoel: The 'M' is ok but for the rest we need to have identifiants and not the english version!

So you need to :
 - Delete the caps of the word
 - Add the translation function on each ID

django/santropolFeast/member/models.py
maximest-pierre: The django doc said that we must call a function that returns the default value.

ref. https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.CharField

Under the section default.

RignonNoel: I've find this example in the same page :
```
class Student(models.Model):
    FRESHMAN = 'FR'
    SOPHOMORE = 'SO'
    JUNIOR = 'JR'
    SENIOR = 'SR'
    YEAR_IN_SCHOOL_CHOICES = (
        (FRESHMAN, 'Freshman'),
        (SOPHOMORE, 'Sophomore'),
        (JUNIOR, 'Junior'),
        (SENIOR, 'Senior'),
    )
    year_in_school = models.CharField(max_length=2,
                                      choices=YEAR_IN_SCHOOL_CHOICES,
                                      default=FRESHMAN)
```

maximest-pierre: I will implement it like the example.

manumilou: @maximest-pierre  Check out the Meal Model, the same thing has already been implemented: https://github.com/savoirfairelinux/santropol-feast/blob/dev/django/santropolFeast/meal/models.py#L23

PR #92, opened 2016-04-26
Fix 30 add referent

### LINKED ISSUE #30: Add referent on client model
We need to have referent information on client
# Proposal

## Tasks

 - Add model 'Referencing'

## Fields

 - ForeignKey to client (Or oneToOne ?)
 - ForeignKey to member (the referent)
 - Date of referencing
 - Description with a TextField (Why is referencing)

maximest-pierre:
fix #30


django/santropolFeast/member/models.py
```diff
@@ -209,3 +209,20 @@ class Meta:
         related_name='allergic_clients',
         blank=True
     )
+
+
+class Referent (models.Model):
```
RignonNoel: Be carefull on the name of the Class. Here i think you describe a reference and not a referent.

maximest-pierre:
What would be an appropriate name for this class.

RignonNoel:
'Reference' maybe? I've not a very good english but i think it's a good candidate

maximest-pierre:
I just check we can change it to referee or referent is accepted.

RignonNoel:
I will speak french for more details on the word:

 - Referee ce traduit par un arbitre, quelqu'un d'officiel, quelqu'un de référence.
 - Referent ce traduit par quelqu'un de référence.

Ici ton model dispose d'un champ ```Member``` et d'un champs ```Client```, il représente donc une référence, comme quand tu écrit un mot pour promouvoir un ancien employé:

 - Ton nom d'employeur ```Member```
 - Le nom de la personne à promouvoir ```Client```
 - La date ou tu as fait le papier

Il ne faut juste pas confondre, le ```référent``` et la ```référence``` qui sont respectivement ```referent``` et ```reference```

maximest-pierre:
Dans l'issues sa indique le nom du model a Referencing. On pourrais mettre sa comme nom. Il faudrait aussi changer les noms de variables pour mieux comprendre le code.

RignonNoel:
Yes it's a good idea.

maximest-pierre:
I am going to change it and then do a commit.

django/santropolFeast/member/models.py
```diff
@@ -209,3 +209,21 @@ class Meta:
...
         blank=True
     )
+
+
+class Referencing (models.Model):
+
+    class Meta:
+        verbose_name_plural = _('referents')
+
+    referent = models.ForeignKey('member.Member',
```
RignonNoel: The ```verbose_name``` must be ```referent``` to have a good display on form and admin form.

django/santropolFeast/member/models.py
```diff
@@ -209,3 +209,21 @@ class Meta:
...
+                                 verbose_name=_('member'))
+
+    client = models.ForeignKey('member.Client',
+                               verbose_name=_('client'))
+
+    referral_reason = models.TextField(
+            verbose_name=_("referral_reason"))
+
+    date = models.DateField(verbose_name=_("referral_date"),
```
RignonNoel: Why you set ```auto_now=False``` ? It can be ok, but i think it's not a bad things if we set a default value.

maximest-pierre: Yeah, we can set to auto_now = True. I think setting to True might be a good idea considering that the referral is almost always the current day.

RignonNoel: Yeah, or create a save function who check if date is Null and set the current date if is True.
I don't know if auto_now overwrite a create form with the date field

maximest-pierre: We can also define default = datetime.date.today() then we can modify it anyway we want.

RignonNoel: Yes, it's ok and more simple.

RignonNoel:
The code is ok for me. You just need to:

 - Delete migrations file and remake a makemigrations (to have only one file)
 - Make all the code in only one commit

RignonNoel:
Thanks!

PR #97, opened 2016-04-28
[WIP] add basic verification for the login page 

### LINKED ISSUE #84: Add validation and error display on the login form

maximest-pierre:
Right now I can make the invalid field red but there is no prompt can somebody with a little more experience with semantic-ui can help me.

fix #84

### DIFF
--- a/django/santropolFeast/page/templates/registration/login.html
+++ b/django/santropolFeast/page/templates/registration/login.html
@@ -17,6 +17,7 @@
 
 {% if not user.username %}
 <form method="post" action='' class="ui large form">{% csrf_token%}
+    <div class="ui error message"></div>
     <div class="ui stacked segment">
         <div class="field">
             <div class="ui left icon input">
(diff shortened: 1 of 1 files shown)


manumilou:
@maximest-pierre I'll have a look.

django/santropolFeast/page/templates/registration/login.html
```diff
@@ -36,6 +36,41 @@ <h2 class="ui teal image header">
         <button class="ui fluid large teal button" type="submit">Login</button>
     </div>
```
manumilou: Add here: <div class="ui error message"></div>
to load error messages.

django/santropolFeast/page/templates/registration/login.html
```diff
@@ -36,6 +36,41 @@ <h2 class="ui teal image header">
...
 </form>
+
```
manumilou: Unnecessary, as semanticUi JS and Jquery are already loaded by the base template.
But be careful to execute the JS on document.ready.

RignonNoel:
**First part :**
Here the JS error message work only when i submit the form. But it's hide directly when i correct the error.

We need instant error, also we can simply write a message of the context processor..

**Second part :**
I think it's important to display error on the top of the form. If we have a bad resolution we can not see the message.

manumilou:
It is all good to me. Just the message on failed authentication that remain to be added.

PR #101, opened 2016-05-02
Fix #53 improve base template [WIP]

### LINKED ISSUE #53: Improve the base template
The frontend framework used is [Semantic UI](http://semantic-ui.com/). It will provide us more advanced  and nicer UI components than Boostrap would do.

The objective here is to develop a top level template that will serve as parent or base layer for most of the application templates.
There is already work done in this area, but it could improved.

The current base template is here: https://github.com/savoirfairelinux/santropol-feast/blob/dev/django/santropolFeast/santropolFeast/templates/base.html

maximest-pierre:
Improve the base templates by adding real names and cleaning the unneed parts.

Fix #53


maximest-pierre:
@manumilou is this template good and are there any changes you would make to it

PR #118, opened 2016-05-10
Improve client admin

cormier:
Add a client formview and listview. Modify Client and Member models to improve their coherence.


django/santropolFeast/member/models.py
```diff
@@ -176,8 +176,8 @@ class Meta:
         verbose_name_plural = _('clients')
 
     billing_address = models.ForeignKey(
-        'member.Address',
-        verbose_name=_('billing_Address')
+        'Address',
+        verbose_name=_('Billing address')
```
RignonNoel: Yeah it's a good point. But we need to keep the 'billing_address' ID to I18N.
It's not an english word, just an ID

django/santropolFeast/member/models.py
```diff
@@ -180,8 +180,8 @@ class Meta:
         verbose_name=_('Billing address')
     )
 
-    facturation = models.CharField(
-        verbose_name=_('facturation_type'),
+    billing_type = models.CharField(
+        verbose_name=_('Billing type'),
```
RignonNoel: Same things. we need to have an ID 'billing_type' and not an english word

RignonNoel:
Nice job, just two point of i18n for me.

Thanks for your contribution :+1:

cormier:
Why use an ID when it is standard practice to use english (or let's say "base language") as a base string? The Django documentation on internationalisation mentions "translation strings", not message ids, and more specifically when it comes to `verbose_name` fields it doesn't say that a "message id" should be use. [1]

https://docs.djangoproject.com/en/1.9/topics/i18n/translation/#model-fields-and-relationships-verbose-name-and-help-text-option-values

I saw that the project uses "message ids". I would consider replacing those with english words.

RignonNoel:
I'm ok with you and i work all the rest of time with english in code.

But here this project work with ID, so if you want change this practice you need to open an issue to discuss with the team. 

I can't accept this PR with english word for the moment.

cormier:
Okay! We will discuss this in #121 and we'll then see what we do with this PR :) Thanks @RignonNoel

RignonNoel:
No problem, you're welcome :)
I close this PR for the moment.

cormier:
@RignonNoel i have rebased this PR on top of #121 

Please consider reopening and merging once we are done with #121 
Thanks!

RignonNoel:
I can't reopen  this PR after a force-push or a re-creation. It's a limitation of Github.
But you can propose a new PR, i'm not close to review your work after acceptation of #123

PR #120, opened 2016-05-10
Add a text area for action related to a client

### LINKED ISSUE #109: Add a text area for action related to a client
There need to be a text area name action added to the model just like the notes class inside the member/models.py.

duckish:
fix #109


RignonNoel:
Seem ok for me but don't respect the DRY principle.
You need to extand the note class for example with multiple heritance.

RignonNoel:
The issue have been change.
So i close this PR

Issue #60, opened 2016-04-21
Check that every string is translatable


manumilou:
For a Python string to be translatable, it needs to be pass through the _() function.

Some strings, that should be translatable, may not respect that. They should be corrected.

RignonNoel:
In template the function is : 
```{% trans 'my text' %}```

maximest-pierre:
We are waiting for a convention for the writing styles. To discuss at the next meeting.

maximest-pierre:
I found some strings that are not translated inside the login.html

RignonNoel:
We change the process : #121
Check every string is translatable with real english version in the code.

PR #123, opened 2016-05-10
I18n do not use keywords

### LINKED ISSUE #121: i18n: use english strings instead of "message ids"
The project uses "message ids" instead of untranslated strings for messages.  This is neither what is recommended by the gettext documentation nor what we generally see in open source projects. When we look at the gettext documentation of po files [1], we see that the message id *is* the untranslated string, not an abstract id. 

For open source projects, the base language is  generally english. This is what we see in the Django project. [2]

I believe that we should avoid abstract message ids because it complicates the translation process. For these reasons I believe that we should:

1. Replace the abstract message ids with english strings
2. Document this decision in the i18n section of the README

References
[1] https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#PO-Files
[2] https://github.com/django/django/blob/ecb59cc6579402b68ddfd4499bf30edacf5963be/django/contrib/admindocs/apps.py#L7

cormier:
This is related to #60. However, these commits do not fix this issue since it does not touch templates.

Fix #121

### DIFF
--- a/django/santropolFeast/member/models.py
+++ b/django/santropolFeast/member/models.py
@@ -38,9 +38,9 @@ class Member(models.Model):
     gender = models.CharField(
         max_length=1,
         choices=(
-            ('M', _('male')),
-            ('F', _('female')),
-            ('U', _('unknown')),
+            ('M', _('Male')),
+            ('F', _('Female')),
+            ('U', _('Unknown')),
         ),
         default='U',
         blank="True",
@@ -84,7 +84,7 @@ class Address(models.Model):
 
     # Member address information
     number = models.PositiveIntegerField(
-        verbose_name=_('street_number')
+        verbose_name=_('street number')
     )
 
     street = models.CharField(
@@ -114,7 +114,7 @@ class Address(models.Model):
     # Montreal postal code look like H3E 1C2
     postal_code = models.CharField(
         max_length=6,
-        verbose_name=_('postal_code')
+        verbose_name=_('postal code')
     )
 
     member = models.ForeignKey(
@@ -135,7 +135,7 @@ class Contact(models.Model):
     type = models.CharField(
         max_length=100,
         choices=CONTACT_TYPE_CHOICES,
-        verbose_name=_('contact_type')
+        verbose_name=_('contact type')
     )
 
     value = models.CharField(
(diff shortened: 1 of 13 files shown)


django/santropolFeast/santropolFeast/settings.py
```diff
@@ -135,13 +135,11 @@
...
 USE_TZ = True
 
-# not import the real utils here! We don't want an infiny loop
-gettext = lambda x: x
 
 # List of supported languages
 LANGUAGES = (
-    ('fr', gettext('French')),
-    ('en', gettext('English')),
+    ('fr', 'French'),
```
RignonNoel: In UI we prefer set language not translated (ex: Français, English, Italiano)
It's easiest to understand for the user.

django/santropolFeast/member/models.py
```diff
@@ -158,30 +158,30 @@ class Client(models.Model):
...
-        (ACTIVE, _('active')),
-        (PAUSED, _('paused')),
-        (STOPNOCONTACT, _('stopnocontact')),
-        (STOPCONTACT, _('stopcontact')),
-        (DECEASED, _('deceased')),
+        (PENDING, _('Pending')),
+        (ACTIVE, _('Active')),
+        (PAUSED, _('Paused')),
+        (STOPNOCONTACT, _('Stop: no contact')),
+        (STOPCONTACT, _('Stop contact')),
```
RignonNoel: It would be nice to homogenize display here. With ```:``` or without ```:```

RignonNoel:
In member/models.py line 9:

```

 HOME = 'Home phone'
 CELL = 'Cell phone'
 WORK = 'Work phone'
 EMAIL = 'Email'
 
 CONTACT_TYPE_CHOICES = (
     (HOME, HOME),
     (CELL, CELL),
     (WORK, WORK),
     (EMAIL, EMAIL),
 )
```

We have no translation on this section.

RignonNoel:
Nice job! :+1: 

I just have a problem with your translation in the feminine gender.

We will soon have a problem with that. I invite you added a translation comment specifying that it is the female that Django treats them apart and so we can translate it twice.

---

EX:
In po file:
```
#: .\path\of\use.html:1
msgctxt "Féminin"
msgid "Ordered"
msgstr "Commandée"

#: .\path\of\use.html:2
msgid "Ordered"
msgstr "Commandé"
```

In template file
```
{% trans "Amount received" context "Féminin" %}
{% trans "Amount received" %}
```

django/santropolFeast/santropolFeast/settings.py
cormier: I agree that it should be done this way. However this is not how how it was done: the `gettext` was only a `lambda x: x` definition. I removed it because it did not add anything.

I will add a commit that put the strings in the original languages.

django/santropolFeast/member/models.py
cormier: Will do. Should we do it with or without the `:`? Maybe someone who knows more about the business logic can help me here.

django/santropolFeast/santropolFeast/settings.py
RignonNoel: Yes, i'm agree it's not your fault. It's just a request :)

cormier:
Thanks for the review! Here is what I intend to do:

* Add contextual markers to the PO files
* settings: put the languages in their original language
* Homogeneize the "stop contact" strings

However I don't think we should do a full review of the i18n in the scope of this PR. The intention of this PR is to make sure that every string that is *already* translated use untranslated english strings instead of message ids.

django/santropolFeast/member/models.py
RignonNoel: We are no business logic about that for the moment (i think).
Personnaly, i like the ```:```

cormier:
I updated this PR to remove the translation of the language names and to homogeneise the use of the colon in the translations. It has also been updated to correct the message-code translations that have been introduced in the project since I first submitted this PR last week.

However, after revewing the documentation on contextual markers I believe that they are out of scope for this PR and should not be included in it. This PR has only one intention: fixing #121 by replacing message codes with message strings. Since there was no contextual markers in the first place, I don't see why they should be added in the scope of this PR.

If you believe that additional tasks should be done to improve the i18n of this project, I invite you to create specific issues.

manumilou:
Agreed. Thx @cormier

PR #108, opened 2016-05-03
[WIP ]Fix 51 implement form views

maximest-pierre:
I might need to make changes to the member models.

fix #51

PR #129, opened 2016-05-13
Fix 51 implement form views

maximest-pierre:
Rebase and clean of PR #108


manumilou:
In the Client model, replace the __str__ function with the following:

```
def __str__(self):
        return "{} {}".format(self.member.firstname, self.member.lastname)
```

PR #134, opened 2016-05-17
Improve member admin

cormier:
Please refer to #118


manumilou:
This PR is outdated. I am closing it, as I understand that a lot of the proposed migrations have already been implemented.
The custom admin interface for the clients is interesting though.

PR #135, opened 2016-05-18
docs : initial visual data model

lamontfr:
This replaces PR https://github.com/savoirfairelinux/santropol-feast/pull/125 which was closed on purpose.

In order to facilitate understanding the overall data model of Santropol Feast, I created one in PlantUML notation and generated the image. The data model includes all the classes and relations currently on the models.py files from all the apps.

I added a data_model folder to docs/. This includes a data model in PlantUML text format (.uml) and a generated image (.png). Then, as suggested by Noël, I added a README.md file that has a URL to the .uml file such that the inline image is dynamically generated.

PlantUML format is a plain text file which allows to see changes made to it using standard code "diffing" tools. See http://plantuml.com


manumilou:
The diagram is great, thanks!
But the link in the README does not give me anything, except an empty diagram on a dark background. Can you arrange this ?

lamontfr:
The README needs the file https://raw.githubusercontent.com/savoirfairelinux/santropol-feast/dev/docs/data_model/current_data_model.uml so you cannot see the image until you merge the PR. May be there is a macro in markdown that can make the URL relative instead of absolute. Otherwise, this is a chicken and egg situation. This is why I also submitted the .png image that I generated locally.

lamontfr:
Thanks for the merge. I looked at the README.md and it still does not display the image. However the same content works in http://markdownlivepreview.com/ . I will find a way to fix it.

PR #138, opened 2016-05-18
Fixed inline diagram image in docs/data_model/README.md

lamontfr:
The dynamic generation of the diagram .png image (by passing the .uml file to the plantuml server) does not work on Github, although it works on http://markdownlivepreview.com/.

Now the README.md shows inline the .png which is in the same folder.


maximest-pierre:
@lamontfr Thank you for the fix.

Issue #154, opened 2016-05-26
Step button needs to link to the right step


maximest-pierre:
### Expected Behaviour

When the user clicks on a step inside the form he should be redirected to that particular step.

### Actual Behaviour

There is no redirecting.

cormier:
Hi! I will start working on this issue

PR #165, opened 2016-05-26
Add links for member create and list to menu

cormier:
Replace stub links with real links!

ref #154


maximest-pierre:
Thank you for the contribution.

PR #171, opened 2016-05-26
Used datetime.date.today as a default value for the Referencing.date field.

charettes:

Used datetime.date.today as a default value for the Referencing.date field.

### Status

Ready

Migration needed (with no resulting schema change)

Using `datetime.date.today()` was setting the default value to the date the server
process was started and also triggered a migration state change every day.

This will silence the fields.W161 check warning.


maximest-pierre:
Thank you for the contribution.

PR #172, opened 2016-05-27
Used setupTestData() to create database test fixtures.

charettes:

Used setupTestData() to create database test fixtures.

### Status

READY

### DIFF
--- a/django/santropolFeast/member/tests.py
+++ b/django/santropolFeast/member/tests.py
@@ -6,7 +6,8 @@ from datetime import date
 
 class MemberTestCase(TestCase):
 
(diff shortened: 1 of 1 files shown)


charettes:
Here's the [documentation](https://docs.djangoproject.com/en/1.9/topics/testing/tools/#django.test.TestCase.setUpTestData) for `setUpTestData()`. The only caveat is when using MySQL/Maria with the MyISAM engine (we're using InnoDB) where `setupTestData()` will be called once per test which is equivalent to what `setUp()` does.

manumilou:
Thanks @charettes

PR #173, opened 2016-05-27
Fix #157 - Fix ClientRestrictionsInformation.status form field

### LINKED ISSUE #157: Status field inside the client form must not be required
### Expected Behaviour

On the Allergies step of the client creation form, the **Meal Status** field should be either Active or not, without any validation problems.

### Actual Behaviour

If the **Meal Status** is set to Inactive, the validation blocks and prevents from getting to the next step.

### Steps to reproduce

* Go the client creation form: _member/create_
* Fill out fields until the _Allergies_ step
* On the _Allergies_ step, set the Meal status to Inactive, and click **Next**
* A validation error occurs, specifying that the status field is mandatory

validation_bug

ellmetha:
*Fixes #157 .*

### DIFF
--- a/django/santropolFeast/member/forms.py
+++ b/django/santropolFeast/member/forms.py
@@ -76,6 +76,7 @@ class ClientRestrictionsInformation(forms.Form):
     status = forms.BooleanField(
         label=_('Active'),
         help_text=_('By default, the client meal status is Pending.'),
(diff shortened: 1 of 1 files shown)


maximest-pierre:
Thanks you for the contribution.

PR #183, opened 2016-05-31
Fix161a - changed models - not tested - for information

lamontfr:
@manumilou 

I still have to check this tonight, but it may be useful as information. **Do not merge this yet !**

There is a bug with client filters, so I have commented it out.

### Status

- [] READY
- [] HOLD
- [X] WIP (Work-In-Progress)


lamontfr:
See my last comment on issue https://github.com/savoirfairelinux/santropol-feast/issues/161

PR #193, opened 2016-06-08
Update the test

maximest-pierre:

* Change the member test to be more compliant with coverall
* Add basic form validation

### Status

- [ ] READY
- [ ] HOLD
- [X] WIP (Work-In-Progress)

@savoirfairelinux/mll

### DIFF
--- a/django/santropolFeast/member/tests.py
+++ b/django/santropolFeast/member/tests.py
@@ -1,4 +1,4 @@
-from django.test import TestCase
+from django.test import TestCase, Client
 from member.models import Member, Client, Note, User, Address, Referencing
 from member.models import Contact, Option, Client_option, Restriction
 from meal.models import Restricted_item
@@ -233,6 +233,7 @@ class RestrictionTestCase(TestCase):
 
 
 class FormTestCase(TestCase):
+
     @classmethod
     def setUpTestData(cls):
         cls.admin = User.objects.create_superuser(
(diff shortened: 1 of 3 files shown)


maximest-pierre:
@RignonNoel @manumilou I cant seem to make the post request pass the correct information. I know that the TestCase disable the crsf protection and the format is good but the data is not saving correctly. Do you guys have any ideas?

hsoft:
@maximest-pierre I am not familiar with the project at all, but what I see is that your tests depend on each other. All your tests depend on `test_acces_to_form` running first. Then, a bunch of tests depend on `test_form_save_data` having ran. On top of that, you log out on tear_down, which I think is ran after each test.

All of that is bad practice.

You should login during a setup phase, your post-post tests should be in their own testcase, with the post operation being part of the setup.

django/santropolFeast/member/tests.py
```diff
@@ -227,3 +230,290 @@ def test_str_includes_all_names(self):
...
+            '/member/create/emergency_contact', emergency_contact_request,
+        )
+
+        print (Member.objects.all)
+        member = Member.objects.get(firstname="Testing")
+
+    def test_member_name(self):
+
+        member = Member.objets.get(firstname="Testing")
+        self.assertTrue(member.firstname, "Testing")
```
hsoft: What you want is `assertEqual()`, not `assertTrue()`.

PR #201, opened 2016-06-16
View order page

### LINKED ISSUE #198: Create a view for order items
### Expected Behaviour

We need a view too see what the item the order contains.

### Actual Behaviour

Curently there is way to see what item the order contains from the front-end.

maximest-pierre:
*Fixes #198 .*

* add url and views to redirect to the view order page
* basic templates for the view order page
* add filtering for order status

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

@savoirfairelinux/mll


maximest-pierre:
@manumilou I think that this pull request is ready. Can you check if there is anything that needs to be modify.

manumilou:
Please rename order_view.html to a simpler view.html

django/santropolFeast/order/templates/list.html
```diff
@@ -29,12 +29,12 @@
   <tbody>
     {% for obj in order %}
       <tr>
-        <td>{{obj.creation_date}}</td>
-        <td>{{obj.delivery_date}}</td>
-        <td>{{obj.client.member}}</td>
+        <td><a href="{% url 'order:view' id=obj.id %}">{{obj.creation_date}}</a></td>
+        <td><a href="{% url 'order:view' id=obj.id %}">{{obj.delivery_date}}</a></td>
+        <td><a href="{% url 'member:view' id=obj.client.id %}">{{obj.client.member}}</a></td>
```
manumilou: The table shoud include a new column with a View action / Button, rahter than links on multiple column.

django/santropolFeast/order/templates/list.html
```diff
@@ -29,12 +29,12 @@
...
         <td>{{obj.status}}</td>
```
manumilou: Display status with: get_status_display.

manumilou:
You could add filter on the order status as well, just like the client list page (the new one :-) ):

django/santropolFeast/order/views.py
```diff
@@ -43,3 +43,9 @@ def get_context_data(self, **kwargs):
         context['get'] = text
 
         return context
+
+
+def show_information(request, id):
+    order = get_object_or_404(Order, pk=id)
+
```
manumilou: Remove this blank line.

django/santropolFeast/order/views.py
```diff
@@ -43,3 +43,9 @@ def get_context_data(self, **kwargs):
...
+    return render(request, 'view_order.html', {'order': order})
```
manumilou: Update this with the new template name.

maximest-pierre:
That would be a cool feature to add.

maximest-pierre:
I got the filter working only on the status and not on get_status_display so it only works with the first letter. I would maybe make a dropdown menu to filter the status.

django/santropolFeast/order/models.py
```diff
@@ -108,6 +114,10 @@ def filter_search(self, queryset, value):
                     if user not in query:
                         query.append(user)
 
+                for user in status:
+                    if user not in query:
+                        query.append(user)
+
```
manumilou: You don't need to use custom filter for the status.
Just use this: https://github.com/savoirfairelinux/santropol-feast/blob/dev/django/santropolFeast/member/models.py#L332

PR #233, opened 2016-06-30
Add <orders> property to <Client>

### LINKED ISSUE #216: Implements a function that gets all orders for a given client
### Expected Behaviour

* A function should be available that returns all orders for a given client

### Actual Behaviour

Nothing

ssirois:
*Fixes #216

### Status

- [X] READY
- [ ] HOLD
- [] WIP (Work-In-Progress)

@savoirfairelinux/mll

Fixes #216

### DIFF
--- a/django/santropolFeast/member/tests.py
+++ b/django/santropolFeast/member/tests.py
@@ -175,7 +175,7 @@ class ClientTestCase(TestCase):
         """Orders associated to a client must be available through client's property"""
         member = Member.objects.get(firstname='Angela')
         angela = Client.objects.get(member=member)
-        self.assertTrue(angela.orders.count(), 1)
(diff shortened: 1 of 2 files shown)


django/santropolFeast/member/tests.py
```diff
@@ -164,6 +171,13 @@ def test_default(self):
         # Delivery type: Ongoing
         self.assertEqual(angela.delivery_type, 'O')
 
+    def test_orders(self):
+        """Orders associated to a client must be available through client's property"""
+        member = Member.objects.get(firstname='Angela')
+        angela = Client.objects.get(member=member)
+        self.assertTrue(angela.orders.count(), 1)
```
maximest-pierre: Change the assertTrue to a assertEqual since its a value and not a condition.

PR #246, opened 2016-07-05
Updated the templates for the delivery/order and add a view

### LINKED ISSUE #224: [CM Step 1] Gets all orders for a given day
### Expected Behaviour

The orders List page should be used actually , as it already lists all orders.
Only a date filter is missing.
It must list:

- Orders of ongoing clients already created (automatically or manually)
- Orders already created in the database for the episodic clients

### Actual Behaviour

An order list without date filter.

maximest-pierre:
*Fixes #224 *

* The order view inside of the delivery/order now shows the order for the current day.

### Status

- [ ] READY
- [ ] HOLD
- [X] WIP (Work-In-Progress)

- [ ] Migration needed

Fill me in ...

The day is date.today() so the filtering is only for the current day. 

@savoirfairelinux/mll


maximest-pierre:
@lamontfr I have a small problem with the form can you take a look at it.

manumilou:
Not valid anymore.

Issue #223, opened 2016-06-29
I should be able to attach an existing member to a client


manumilou:
### Expected Behaviour

During the client creation form, an emergency contact and a referrent are attached to the client.
I should able to select an existing member, or create a new one if he is not in the database yet.

Semantic-UI provides a nice way of implementing this behaviour. 

![select_or_create](https://cloud.githubusercontent.com/assets/212768/16468880/05863994-3e1c-11e6-8290-92f3e472c4aa.png)

### Actual Behaviour

A new member is created every time. There is no possibility to lookup up for an existing member.

PR #247, opened 2016-07-10
Add the possibility to attach an existing member to the client.

erozqba:
*Fixes issue #223 .*

* Update the tests because in a previous commit I had used assertTrue instead of assterEqual
* Add a route and a view for making ajax request in order to
find the existing members.
* Update the templates to add the new input to search existing
members and the JavaScript code needed.
* Update the view ClientWizard to add the existing member to the
client when needed, also split the long save method into several
logic methods that are reused in the save method.

 * Important change : Right now I have decided that if the
client information and the payment information is the same
(First Name and Last Name) only one member is created in the
DataBase, because if not we end with lots of members that basically
have the same information. Is the same with the payment information
and the referent information, maybe the person referencing the
client is the same that will pay for it, and right now we create
two differents members in the DataBase. I'm not sure how this should
be handled and I will be happy to change this into the right way.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

- [ ] Migration needed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,5 @@ ENV/
 # Rope project settings
 .ropeproject
 
+# IDE project settings
+.idea
--- a/django/santropolFeast/member/forms.py
+++ b/django/santropolFeast/member/forms.py
@@ -243,8 +243,6 @@ class ClientPaymentInformation(MemberForm):
                 field_data = cleaned_data.get(field)
                 if not field_data:
                     self.add_error(field, msg)
-
-
         return cleaned_data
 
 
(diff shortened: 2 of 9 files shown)


maximest-pierre:
Everything looks good except the coverage has taken a hit. Can you look at improving the coverage.

Thank you.

erozqba:
@maximest-pierre I will look into it when I have the time.

erozqba:
Hi @maximest-pierre, I have looked into the coverage and increased it for the current work, so it should be ready to be merged.

I find out that the coverage can be increased in other areas, I will try to look at it when I have more free time.

maximest-pierre:
It looks good to me. Thanks for the PR.

PR #254, opened 2016-07-18
Fix the problem with the templates #253

### LINKED ISSUE #253: Changing the client id to pk in templates
### Expected Behaviour

The templates must render a link with the pk and not the id.

### Actual Behaviour

Throws an error with the template.

### Steps to reproduce

* Go to any templates that has a reference to the member view
* The templating engine will throw and error.

maximest-pierre:
*Fixes #253 *

* Fix the problem with the reverse lookup (The id was change by an early commit to pk). 

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

- [ ] Migration needed

@savoirfairelinux/mll

### DIFF
--- a/django/santropolFeast/member/templates/client/list.html
+++ b/django/santropolFeast/member/templates/client/list.html
@@ -122,7 +122,7 @@
     <tbody>
         {% for obj in clients %}
         <tr>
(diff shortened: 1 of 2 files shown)


erozqba:
Thanks, I miss this in my last commit. Sorry!

PR #258, opened 2016-07-23
Add new commands to generate orders using factory-boy

erozqba:

* Update ClientFactory to add new fields like the json meals_default_week
and create ReferencingFactory
* Reuse the same constants COMPONENT_GROUP_CHOICES and DAYS_OF_WEEK in all
classes and remove duplication
* Update the command createclients
* Add the commands createorders and generateorders. The command createorders
will create new client and orders with ramdom data. The command generateorders
will create orders for the existing clients preferences of meal_default_week.
* Update tests files to the previous changes made.
* Update dataload.py because the component groups names has changed.

### Status

- [X] READY

### DIFF
--- a/django/santropolFeast/meal/models.py
+++ b/django/santropolFeast/meal/models.py
@@ -18,7 +18,7 @@ INGREDIENT_GROUP_CHOICES = (
     ('dairy', _('Dairy')),
     ('fish', _('Fish')),
     ('seafood', _('Seafood')),
-    ('veggies and fruits', _('Veggies and fruits')),
+    ('veggies_and_fruits', _('Veggies and fruits')),
     ('legumineuse', _('Legumineuse')),
     ('grains', _('Grains')),
     ('fresh_herbs', _('Fresh herbs')),
--- a/django/santropolFeast/meal/models.py
+++ b/django/santropolFeast/meal/models.py
@@ -2,11 +2,11 @@ from django.db import models
 from django.utils.translation import ugettext_lazy as _
 
 COMPONENT_GROUP_CHOICES = (
-    ('main dish', _('Main dish')),
+    ('main_dish', _('Main Dish')),
     ('dessert', _('Dessert')),
-    ('diabetic dessert', _('Diabetic dessert')),
-    ('fruit salad', _('Fruit salad')),
-    ('green salad', _('Green salad')),
+    ('diabetic', _('Diabetic Dessert')),
+    ('fruit_salad', _('Fruit Salad')),
+    ('green_salad', _('Green Salad')),
     ('pudding', _('Pudding')),
     ('compote', _('Compote')),
 )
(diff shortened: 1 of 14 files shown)


lamontfr:
@erozqba Thanks for your great contributions, For this one, in file django/santropolFeast/order/models.py in line 150 you changed
 **order = Order.objects.get(client=client, delivery_date=delivery_date)**
for
**Order.objects.get(client=client, delivery_date=delivery_date)**
However, the **order =** is necessary because when the **try** does not raise an exception,  in line 188 we will need **order** to have been previously defined
**oi = Order_item(order=order, ...**

django/santropolFeast/order/models.py
```diff
@@ -147,8 +146,7 @@ def create_orders_on_defaults(creation_date, delivery_date, clients):
             if free_side_dish_qty == 0:
                 continue  # No meal for client on this day
             try:
-                order = Order.objects.get(
-                    client=client, delivery_date=delivery_date)
+                Order.objects.get(client=client, delivery_date=delivery_date)
                 # order already created, skip order items creation
                 # (if want to replace, must be deleted first)
                 continue
```
erozqba: > @erozqba Thanks for your great contributions, For this one, in file django/santropolFeast/order/models.py in line 150 you changed
> order = Order.objects.get(client=client, delivery_date=delivery_date)
> for Order.objects.get(client=client, delivery_date=delivery_date)
> However, the order = is necessary because when the try does not raise an exception, in line 188 we will need order to have been previously defined
> oi = Order_item(order=order, ...
> 

Hi @lamontfr, thanks a lot for your revision, I don't find your comment here in Github, so I have pasted the content of the mail to have the proper context for my question. I change the line 150 because all this code is inside a _**for**_ loop and in line 152, inside the **_try_** there is a _**continue**_, so if an exception is not raised the execution of the code will follow with the next iteration of the loop and will only get to the line 186 if an exception is raised. Have I missed something there?

lamontfr: Hi Ernesto,

You are completely right. After I made the comment, I checked the code
again and realized that I was wrong, then I deleted the comment. I
should have instead edited the comment so say "Ignore it". Sorry about
that.

By the way, are you in Montreal ? I would like to talk with you about
how we can plan our respective contributions to SousChef. Please let
me know when would be a good time to call you.

-Francois.

erozqba: Hi @lamontfr,
Yes, I'm in Montreal.
You can send me your contact information to my email erozqba@gmail.com and when would be a good time to call you.

django/santropolFeast/meal/models.py
```diff
@@ -21,10 +21,10 @@
     ('veggies and fruits', _('Veggies and fruits')),
```
lamontfr: @erozqba **('veggies and fruits', _('Veggies and fruits')),** should be **('veggies_and_fruits', _('Veggies and fruits')),**

django/santropolFeast/order/models.py
```diff
@@ -182,9 +180,9 @@ def create_orders_on_defaults(creation_date, delivery_date, clients):
                     else:
                         unit_price = side_price
                         while free_side_dish_qty > 0 and default_qty > 0:
-                            free_side_dish_qty = free_side_dish_qty - 1
-                            default_qty = default_qty - 1
-                            free_quantity = free_quantity + 1
+                            free_side_dish_qty -= - 1
```
lamontfr: @erozqba **free_side_dish_qty -= - 1** should be **free_side_dish_qty -= 1**

lamontfr:
@manumilou This PR 258 seems great to me. Please merge it when possible. I am waiting for this PR 258 to add my PR for the improved Kitchen Count Report.

django/santropolFeast/meal/models.py
erozqba: :+1:

django/santropolFeast/order/models.py
erozqba: :+1:

PR #259, opened 2016-07-25
Use only active clients when generating orders

manumilou:

* Use active clients only when generating orders

### Status

- [X] READY

@savoirfairelinux/mll

Issue #258


erozqba:
:+1:

PR #306, opened 2016-08-04
Update Order

erozqba:
*Fixes # .*

* Add the form to update orders
* Use django-extra-views dependency
* Use django-dynamic-formset dependency
* Use CBV instead FBV

* TODO: Add tests, I will do it as soon I have some time.

### Status

- [X] READY

- [ ] Migration needed


erozqba:
Hi @manumilou, I thinks this will do it for the update order form. We can look into change the widget too.

manumilou:
It seems to break the UpdateOrderStatus feature.

erozqba:
Hi @manumilou It should work now the update order status

PR #311, opened 2016-08-05
Add tests for the create and update orders forms.

erozqba:
*Fixes: The code for the orders form without tests.*

* Add the tests for the create and update orders forms.
* Improvements in the templates for list, view, create and update orders.
* Small fix in the member test
* Small change in the views of order to set the list of objects like orders instead of order
* Small change in main.css to improve the display of errors

### Status

- [X] READY

- [ ] Migration needed

### DIFF
--- a/django/santropolFeast/member/tests.py
+++ b/django/santropolFeast/member/tests.py
@@ -362,9 +362,8 @@ class ClientAvoidComponentTestCase(TestCase):
 
 class FormTestCase(TestCase):
 
-    @classmethod
-    def setUpTestData(cls):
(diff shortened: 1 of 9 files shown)


django/santropolFeast/member/tests.py
```diff
@@ -918,6 +920,7 @@ def test_form_validate_data(self):
         self._test_step_dietary_restriction_without_errors()
         self._test_step_emergency_contact_with_errors()
         self._test_step_emergency_contact_without_errors()
+        self._logout()
```
manumilou: @erozqba Why not in setUp() and tearDown function ?

Issue #357, opened 2016-08-19
Convert "through" models to ManyToManyFields


cormier:
Instead of using ManyToManyFields for 1..n relationships, "through" models are defined. This results in added complexity in writing code that queries the database.

Here is an example from the source:

Clients have ingredients that they wish to avoid. Those are defined by the model `Client_avoid_ingredients` in https://github.com/savoirfairelinux/sous-chef/blob/c3e1791876bd8058c82cadc05cf070604351b40d/django/santropolFeast/member/models.py#L749.

If one wants to have the list of the ingredients a particular client wants to avoid, the following code has to be used:

``` python
client_avoid_ingredients = Client_avoid_ingredients(client=client)
ingredients = [ingredient for client_avoid_ingredient.ingredient in client_avoid_ingredients]
```

With a ManyToManyField, it would be possible to have this list directly:

``` python
ingredients = client.ingredients_to_avoid.all()
```

If one wants to have the list of clients who wants to avoid a specific ingredient, the following code has to be used:

``` python
client_avoid_ingredients = Client_avoid_ingredients(ingredient=ingredient)
clients = [client for client_avoid_ingredient.client in client_avoid_ingredients]
```

With a ManyToManyField, this list would be more easily accessible:

``` python
clients = Client.objects.filter(ingredients_to_avoid=ingredient)
```

The code, in its current form, acknowledges the need for an easy way to get the list of ingredients in the context of a client. This is why the `ingredients_to_avoid` property has been introduced:

https://github.com/savoirfairelinux/sous-chef/blob/c3e1791876bd8058c82cadc05cf070604351b40d/django/santropolFeast/member/models.py#L466

However, defining `ingredients_to_avoid` as a M2M field would provide a much easier way to access it. Also, it would reduce the complexity of the code by removing the `Client_avoid_ingredient` model.

Furthermore, properly defining the relationships between the models could help us work towards implementing a pure Django solution to #339
## Todo

Here is the list of the models that serve no other purpose than implementing a 1..n relationship and that should be converted to Django m2m fields:
- [ ] Client_avoid_ingredient
- [ ] Client_avoid_component
- [ ] Client_option
- [ ] Restriction
- [ ] Component_ingredient
- [ ] Incompatibility
- [ ] Menu_component
- [ ] Order_item
- [ ] OrderBilling

lamontfr:
[...]

PR #359, opened 2016-08-20
Replace dataload and dataexec with initial_data fixture

cormier:
This PR replaces custom data loading script with their equivalent Django fixture. This is necessary in order to make progress on #357


cormier:
Superseded by #362

PR #362, opened 2016-08-21
Replace dataload and dataexec with initial_data fixture, original ide…

erozqba:
Replace dataload and dataexec with initial_data fixture, original idea @cormier

Cormier has proposed to use a fixture instead of dataexec and dataload, and
that is a good idea to me. But I prefer not to delete the dataexec and dataload
files, because the fixtures created was a dump of the database after populate
the database with dataexec and dataload, so it will be easier in the future
to made modifications in dataload and use it to generate a new version of
the fixture initial_data.json.

Ref: issue #357  pr #359


cormier:
I believe that we should remove dataexec and dataload from the project because they need to be maintained every time we modify the models and because the Django framework already provides features for creating and managing initial data.

If we need to update the initial data fixture, we can do so by modifying the data directly in the Django admin and calling dumpdata. This has the advantage of freeing us from the maintenance of the custom dataload scripts.

If we decide on keeping the custom dataload scripts, I think we should avoid introducing fixtures, as we will then have two sources of initial data to maintain and that they will most likely end up being out of sync.

erozqba:
You are totally right!

cormier:
I was working on my old PR at this moment :smile: 

I think we need to do the following things:

1) export only the apps of the project
2) change the loaddata calls in the tests to `call_command` calls

cormier:
I looked at your commit and using  `fixtures = ['initial_data']` instead of call_command is even better! :smile:

lamontfr:
@erozqba @cormier Please note that the dataload.py file is generated automatically from a spreadsheet that I use to build initial realistic data. This is temporary until we convert production data from the FeastDB system at Roulant. I have more data to generate for the recipes (component ingredients) and restricted items using the spreadsheet which is much faster than adding data using the admin. It would be quite a bit of work in the spreadsheet to generate json instead of dataload.py. I will have now to use your first approach which is to : run dataexec then dumpdata to make the fixture.

cormier:
@lamontfr, thanks for the clarifications!

It is difficult to help with the spreadsheet issue because it doesn't seem to be in the repository.

Is there a lot of model logic coded into that spreadsheet? What do you recommend? Should we just ignore the dataload script when we make changes to the models and assume you will modify the spreadsheet accordingly? Should we try to turn this spreadsheet into an import script that relies on our model definitions to avoid model logic duplication?

lamontfr:
@cormier The spreadsheet and dataload.py are temporary until we convert real production data. I will simply use them locally on my machine to update parts of the fixture file as needed.

cormier:
Okay! I just wanted to be sure we don't break anything. :smile: 
So I guess we are good to go!

PR #366, opened 2016-08-23
Client: replace properties with m2m fields

cormier:
## Context

Replace getter properties with m2m fields as discussed in #357 

## Changes to the models

* Replace `ingredients_to_avoid` with a proper m2m definition (ref #357)
* Replace `components_to_avoid` with a proper m2m definition (ref #357)
* Add `options` m2m relation (ref #357)
* Keep the `Client_avoid_ingredient` and `Client_avoid_component` model definition to keep compatibility with custom dataload script (ref #362)

## Changes to the admin

Replace `Client_avoid_component` and `Client_avoid_ingredient` forms with tabular inlines in the client form


erozqba:
:+1:

manumilou:
I guess we could remove the intermediate models as well ? At least the ones that do not provide extra fields ?

cormier:
@manumilou that is the end goal, but I kept them there to assure compatiblity with with the dataload / dataexec scripts (#362)

PR #367, opened 2016-08-23
Feature/scheduling client status updates

### LINKED ISSUE #313: Be able to schedule the client delivery status changes
### Expected Behaviour

I should be able to change a client meal's status at a date further in time.
Example: _Mme Michu will pause her deliveries from 1st september to 15th september._

### Actual Behaviour

The status can be changed and is immediately effective.

philippemouchel:
*Fixes #313.*

* Edit ClientStatusScheduler to provide a default list order
* Add a magic method to list all client scheduled status
* Define a filter form to display all scheduled status for a client

### Status

- [ ] READY
- [ ] HOLD
- [X] WIP (Work-In-Progress)

- [ ] Migration needed

@savoirfairelinux/mll


django/santropolFeast/member/templates/client/view/status.html
```diff
@@ -0,0 +1,69 @@
...
+            </div>
+        </div>
+    </h2>
+    <div class="ui divider"></div>
+
+    <form action="" method="GET" class="ui form">
+        <div class="inline fields">
+            <div class="field">
+                <label>{% trans 'Operation status' %}</label>
+                {{ filter.form.operation_status }}
```
erozqba: You should be able to use ` <label>{% filter.form.operation_status.label %}</label>` here.

django/santropolFeast/member/views.py
```diff
@@ -632,6 +632,34 @@ def get_context_data(self, **kwargs):
...
+class ClientStatusView(generic.DetailView):
+    model = Client
+    template_name = 'client/view/status.html'
+
+    # DON'T KNOW WHAT THIS METHOD IS USED FOR
+    # STATUS PAGE IS OK WITH OR WITHOUT..
+    # @method_decorator(login_required)
+    # def dispatch(self, *args, **kwargs):
+    #     return super(ClientStatusView, self).dispatch(*args, **kwargs)
+
```
erozqba: `@method_decorator(login_required)` will check that the user is logged to be able to see the page, so if the URL should be visible to logged users only, we need to use this.

philippemouchel:
Reste un pb de JS : le dropdown pour modifier/scheduler le statut du client ne marche pas sur la page de liste des statuts : member/view/1/status
Alors qu'il fonctionne sur toutes les autres !!

django/santropolFeast/member/views.py
philippemouchel: Merci @erozqba, je retire les commentaires et je repushe

django/santropolFeast/member/views.py
```diff
@@ -632,6 +632,34 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class ClientStatusView(generic.DetailView):
+    model = Client
```
erozqba: Why not use `generic.ListView` https://docs.djangoproject.com/en/1.10/ref/class-based-views/generic-display/#listview
you are showing a page with the lists of `client_statuses`. 

> class django.views.generic.list.ListView¶
A page representing a list of objects.

> While this view is executing, self.object_list will contain the list of objects (usually, but not necessarily a queryset) that the view is operating upon.

philippemouchel: Because it's basically a copy/paste from ClientNotesView ;)

django/santropolFeast/member/models.py
```diff
@@ -489,6 +489,13 @@ def notes(self):
         return self.client_notes.all()
 
     @property
+    def scheduled_statuses(self):
+        """
+        Returns scheduled status associated to this client
+        """
+        return self.client_scheduled_statuses.all()
+
```
erozqba: @philippemouchel if you change the declaration in the model `ClientScheduledStatus` from 
```
client = models.ForeignKey(
        Client,
        on_delete=models.CASCADE,
        related_name='client_scheduled_statuses'
    )
```
to 
```
client = models.ForeignKey(
        Client,
        on_delete=models.CASCADE,
        related_name='scheduled_statuses'
    )
```

you don't really need to define this property, it will remain the same in the others parts of the code, you could access to the client statuses using `self.scheduled_statuses.all()` in the class or `client.scheduled_statuses.all()` in the object. You also will be able to filter the status, I think it will remain more flexible.

philippemouchel: ?? I don't understand, you simply ask me to rename a «method»..
OK, name is shorter, PEP8 compliant...
I already use this related name in a property method of Client model (l.492)

django/santropolFeast/member/templates/client/view/status.html
philippemouchel: `<label>{{ filter.form.operation_status.label }}</label>` is OK..
Error with `{% .. %}`

erozqba: @philippemouchel copy/paste error, sorry. {% %} is for call to function tags like 'trans', {{ }} is for print values..

philippemouchel:
Same feature in #370 PR !

PR #369, opened 2016-08-24
There are a regression added in the rename_app PR that made the JS files

erozqba:
*Fixes a regression in the calendars popups .*

* I have moved the js/css files from the calendar module to a different folder to be consistent with the other modules used.

I will propose others changes in other PR to use gulp and have a place to the frontend development, with minification, etc, because we have too much of JS in the Django templates right now.

### Status

- [X] READY


manumilou:
The regression has already been fixed in #343 .
I guess the question is either we organize the assets by module or by type. The first step should be, as you said, to extract  inline JS.

PR #390, opened 2016-08-25
Fixes CSV and PDF export links within order list page

### LINKED ISSUE #373: The orders CSV export is broken
### Expected Behaviour

On the orders list page (/orders/list), I should be able to export the current queryset to a CSV file, just as for the clients.

The informations to export are, at least:

* The order ID
* The client name
* The order's amount
* The order's status

### Actual Behaviour

The export route does not exist (404 error).

capture d ecran de 2016-08-25 01-04-35

### Steps to reproduce

* Click on the **Orders** menu item
* At the bottom of the table, click on **Export > CSV**

madmath:
*Fixes #373 by fixing the export links*

It used to only work if you had a query string (i.e. had filtered the order list). Will now work without an existing query string as well (URL becomes "/order/list/?&format=csv", which works).

* Change the way the URL is built in the template, to preserve the query string and append the "format" param.

### Status

- [X ] READY

- [ ] Migration needed

### DIFF
--- a/src/sous-chef/templates/system/_button_export.html
+++ b/src/sous-chef/templates/system/_button_export.html
@@ -4,8 +4,8 @@
         <i class="dropdown icon"></i>
                 <div class="menu transition hidden">
                     <div class="header">Format</div>
-                    <div class="item"><a href="{{request.get_full_path}}&format=csv">CSV</a></div>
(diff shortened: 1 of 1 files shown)


manumilou:
Thank you @madmath ! It does work properly now.

madmath:
Sorry, there is no migration needed even though it might have seemed this way in my description.

PR #397, opened 2016-08-25
Display 'Active','Pending',etc instead of 'A','P' in client_export.csv

### LINKED ISSUE #375: The client CSV export should use the status display name
### Expected Behaviour

The client's status should be displayed in full text in the CSV export.

### Actual Behaviour

The client's status is displayed as a code in the CSV export.

### Steps to reproduce

* Click on the **Clients** or the **Orders** menu items. 
* For the clients list, you need to toggle the table view, instead of the Card view
* Click on the dropdown icon on the Export button
* Open the CSV file and look at the status column. Only a letter is displayed.

amedeedaboville:
*Fixes #375*

Changes the ClientExportCSV function to show the human readable client status instead of the 1 character version stored in the database.

This means that the client_export.csv file, under 'Client Status' will have eg, 'Active', or 'Pending' instead of 'A' or 'P'.

### Status

- [X] READY

- [X] No migration needed

### DIFF
--- a/src/member/views.py
+++ b/src/member/views.py
@@ -667,7 +667,7 @@ def ExportCSV(self, queryset):
             obj.id,
             obj.member.firstname,
             obj.member.lastname,
-            obj.status,
+            obj.get_status_display(),
             obj.alert,
             obj.gender,
             obj.birthdate,


amedeedaboville:
I noticed I can also fix the the Delivery Type from 'E' and 'O' to 'Episodic' and 'Ongoing'. Would this be good?

manumilou:
One liner, beautiful :+1:

Issue #389, opened 2016-08-25
Default sensitive values should be provided on the Staff notes list


manumilou:
### Expected Behaviour

By default, only the unread notes should be displayed.
### Actual Behaviour

All the notes are displayed by default.
### Steps to reproduce

_Click on the *_Staff Notes*\* menu item  

Related issues: #379 , #378

Issue #379, opened 2016-08-25
Default sensitive values should be provided on the clients list


manumilou:
### Expected Behaviour

By default, the active clients should be displayed.

![capture d ecran de 2016-08-25 01-59-00](https://cloud.githubusercontent.com/assets/212768/17958399/9670c45c-6a67-11e6-9876-df9ec4735c3d.png)

### Actual Behaviour

No filters.

![capture d ecran de 2016-08-25 01-59-03](https://cloud.githubusercontent.com/assets/212768/17958404/9d17b892-6a67-11e6-922a-e022935702ae.png)

### Steps to reproduce

* Click on the **Clients** menu item

Issue #378, opened 2016-08-25
Default sensitive values should be provided on the order list


manumilou:
### Expected Behaviour

By default, the orders of the current day, with the Ordered status, should be displayed.

![capture d ecran de 2016-08-25 01-55-11](https://cloud.githubusercontent.com/assets/212768/17958345/11aab1ce-6a67-11e6-83aa-481c0f6e9323.png)
### Actual Behaviour

No filters.

![capture d ecran de 2016-08-25 01-55-22](https://cloud.githubusercontent.com/assets/212768/17958360/28f1063a-6a67-11e6-8f93-6d548f6f6ecc.png)
### Steps to reproduce
- Click on the **Orders** menu item

Related issues: #379 , #389

PR #412, opened 2016-08-26
Implement default filters in main menu for clients (#379), orders (#3…

benoitg:
- Implement default filters in main menu for clients (#379), orders (#378), and staff notes (#389)
### Status
- [ ] READY
- [ ] HOLD
- [X] WIP (Work-In-Progress)
- [ ] Migration needed

@savoirfairelinux/mll

### DIFF
--- a/src/sous-chef/templates/system/menu.html
+++ b/src/sous-chef/templates/system/menu.html
@@ -18,7 +18,7 @@
 </div>
 {% endif %}
 <div class="item">
-    <a class="item" href="{% url 'member:list'%}">{% trans 'Clients' %}
+    <a class="item" href="{% url 'member:list'%}?status={{ CLIENT_FILTER_DEFAULT_STATUS | urlencode }}">{% trans 'Clients' %}
         <div class="ui label">{{ CLIENTS_TOTAL }}</div>
     </a>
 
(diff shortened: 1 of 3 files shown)


manumilou:
A better way would be to use the `get_default_ops_value` function in the view.

For an example., have a look there: https://github.com/savoirfairelinux/sous-chef/blob/dev/src/member/views.py#L856
The filter value is default to TOBEPROCESSED value, if no get params are already set.

benoitg:
@manumilou While that would also fulfil the requirements as written, I respectfully disagree it would be a better way.  Relying on get_default_ops_value would preclude having two menu items target two different filter configuration on the same model in the future (or at least force using two different mechanism to do so).  On a more general principle, I believe having restrictive filters in default configuration should be used sparingly (for example excuding tombstoned or expired data and things of that nature), if only because it can affect a lot of unrelated tests.

Normally I'd expect the menu view to interact with the relevant model views to inject the requested configuration.  Since we don't have a menu view, the best I found following existing pattern in the menu code, it to add it to the context manager.

Anyway, do feel free to change it if you really believe it's better, I won't be insulted.

src/sous-chef/templates/system/menu.html
```diff
@@ -30,10 +30,10 @@ <h4 class="ui header name">
             </div>
         </form>
     </div>
-    <a class="item" href="{% url 'order:list' %}">{% trans 'Orders' %}
+    <a class="item" href="{% url 'order:list' %}?status={{ ORDER_FILTER_DEFAULT_STATUS | urlencode }}&delivery_date={{ ORDER_FILTER_DEFAULT_DATE | date:'SHORT_DATE_FORMAT' | urlencode }}">{% trans 'Orders' %}
```
manumilou: The date filter does not work with the formatting you provide.

`ORDER_FILTER_DEFAULT_DATE | date:'Y-m-d' `

manumilou:
@benoitg Sorry for the delay ... I think you are actually right. The menu item should provide the query parameters. I just added a comment to fix the date format passed as parameter.

PR #415, opened 2016-08-26
Supply proper redirect ('next') value when deleting an order

### LINKED ISSUE #411: Deleting an order from the order list resets the query params
### Expected Behaviour

Query params should be kept after deletion.

### Actual Behaviour

Query params are reset.

### Steps to reproduce

* Go the order list
* Select a filter and click search
* Delete any order 
* The search filter has been reset

madmath:
*Fixes #411 by providing a 'next' url param.*

* The template for delete confirmation will now POST to a URL and provide its next destination, which is the order list with the same current set of URL params.
  * This way, when deleting an order, the next page will contain the same search filters. 

### Status

- [X] READY

Fill me in ...

In the previous situation, at the point where the deletion was confirmed in DeleteOrder, it was difficult to get back to the previous set of URL parameters. A common pattern in Django is to provide a 'next' parameter, which we use here to follow the URL to the filtered order list.

@savoirfairelinux/mll

### DIFF
--- a/src/order/templates/order_confirm_delete.html
+++ b/src/order/templates/order_confirm_delete.html
@@ -9,7 +9,7 @@
       <p>{% trans 'This order will be deleted and all the info will be lost, are you sure?' %}</p>
   </div>
   <div class="actions">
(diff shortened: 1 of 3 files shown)


manumilou:
Works as expected :+1:

PR #441, opened 2016-09-01
Fix regression in order edit page.

### LINKED ISSUE #440: Order update form is broken
### Expected Behaviour

We should be able to edit the order and its items.

### Actual Behaviour

Crash on saving the order.

capture d ecran de 2016-09-02 11-51-56

### Steps to reproduce

* Edit an order
* Change information
* Save

erozqba:

* Fix regression in the update order page and form.

### Status

- [X] READY

### How to verify this change

* Edit an order
* Change information
* Save
* Verify that the order has been updated with you changes

### New translatable strings

### Additional notes

Ref: issue #440


manumilou:
The update does not crash anymore. But it is still impossible to remove an order item. Clicking on the button _Remove order items_ removes all items from the interface, but it is not saved properly. Can you reproduce that ?

erozqba:
@manumilou check again!

Issue #435, opened 2016-09-01
Implements a route optimization algorithm


manumilou:
### Expected Behaviour

On the last step of the Kitchen Count process, delivery routes should be optimized by default.

### Actual Behaviour

A default order that does not really make sense is provided. Then, it should be re-ordered manually.

PR #445, opened 2016-09-02
Optimize initial route

larose:
Part of #435

### DIFF
--- a/src/delivery/fixtures/delivery_route_data.json
+++ b/src/delivery/fixtures/delivery_route_data.json
@@ -119,8 +119,8 @@
         "floor": 10,
         "city": "Montreal",
         "postal_code": "H1A2B3",
-        "longitude": null,
-        "latitude": null,
+        "longitude": 4,
+        "latitude": 4,
         "distance": null
     }
 },
@@ -134,8 +134,8 @@
         "floor": null,
         "city": "Montreal",
         "postal_code": "H3D4G6",
-        "longitude": null,
-        "latitude": null,
+        "longitude": 6,
+        "latitude": 6,
         "distance": null
     }
 },
@@ -149,8 +149,8 @@
         "floor": null,
         "city": "Montreal",
         "postal_code": "H4E3WS",
-        "longitude": null,
-        "latitude": null,
+        "longitude": 5,
+        "latitude": 5,
         "distance": null
     }
 },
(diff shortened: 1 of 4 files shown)


larose:
This pull request conflicts with the changes of https://github.com/savoirfairelinux/sous-chef/pull/443 in `src/delivery.py:DailyOrders`.

This pull request optimizes the route returned to the user based on the position of the clients. When I wrote the algorithm, the route returned to the user was random, now it is a route previously saved.

@lamontfr, @manumilou do you still need this optimization. If yes, how?

lamontfr:
Mathieu,

I will check this tomorrow and let you know.

-François.

lamontfr:
@larose @manumilou I think that Mathieu's route optimization is an important improvement and is definitely required.

When I worked on PR #443, I was not aware of your work and here were my assumptions : 
- The user will choose a route, see the list of orders to be delivered on this day, then *manually* adjust the sequence by looking at the points on the map and finally save this sequence using the existing button.
- Most clients receiving meals are regular, such that for a given day of the week the list is similar except for a few additions.

So the code I wrote does this : 
- When the user chooses a route, we load all the orders for the day, we look for a saved sequence from the same weekday of the previous week and if it exists, we use the latter to sort the clients that correspond, then we put the new clients at the end.

Now that you wrote a path optimisation algorithm (TSP) , we have two choices : 

1- When the user chooses a route, we load all the orders for the day in the sequence saved from last week (if any).  Then when the user clicks "Optimize" we use your TSP optimisation to find the best sequence then display these on the map.  The user can then make some manual changes and save this sequence, then print/display the delivery Route Sheet.

2- When the user chooses a route, we load all the orders for the day then we use your TSP right away to find the best sequence then display these on the map.  The user can then make some manual changes then print/display the delivery Route Sheet *without* saving the sequence.

I am open to any of these choices. Let me know what you think.

manumilou:
Thanks @lamontfr for the recap of the situation.

I would rather go for the solution 2. For two reasons:

* It is simpler if the user does not need to click _Save_ before Printing
* It is more logical to use a default behaviour based on the TSP algorithm proposed by @larose, rather than using the order of the last week, which most probably will have to be different

larose:
Ready to be merged! Thanks @lamontfr and @manumilou!

manumilou:
Thanks @larose :+1:

PR #448, opened 2016-09-05
Fix447 Load all ingredients, recipes and restriction data

### LINKED ISSUE #447: Load Ingredients, Main dishes, Recipes, Restricted items, Incompatibilities
**Need script to load data prepared by Katrina on 2016-07-01**

lamontfr:

* Rename existing ingredients, components and restricted_items by prefixing "OLD" to avoid loss of integrity.
* Load new data obtained from Katrina on 2016-07-01 for all ingredients, main dishes and their recipes, restricted_items and their ingredients.

### Status

- [X] READY

- Backup existing database
- run python dataexec.py (this will import dataload.py and call insert_all())
- see progress messages on console until "END dataload".

### Additional notes

*The 2 files you need are in the last commit 	da39f0f.  Sorry about the extra commits from my previous PR which is still open. My laptop battery is very low and my power supply is broken so I have to stop now.*


manumilou:
@lamontfr Don't we use the fixtures now for that ? If yes, it's ok to still use the dataload script to generate, but then you should dump the data into fixtures.

Other question: is there a PR between #443 and #448 to merge first ?

manumilou:
I will update the fixtures after the merge.

PR #454, opened 2016-09-10
Continue dockerize the application.

### LINKED ISSUE #439: Discussions around deployment solutions
### Expected Behaviour

We should have a rather generic solution to deploy the application.
Even though the application won't be on internet @Santropool, it may be different for other *popottes*.

### Actual Behaviour

Dockerized application, run with a development server.

erozqba:

* Add a Nginx docker container.
* Add gunicorn as WSGI server.
* Add differents docker-compose configurations files for DEV and PROD env.
* Add a volume to made database files persistent so it could backup it later.
* Update INSTALL.md file to explain how to access to the application using the
Nginx server and how to run the Django development server.
* Remove some annoying "print()" in the tests.
* Update the .po files.

### Status

- [X] READY

### How to verify this change

* Stop all your containers (ctrl+c) in the terminal where docker-compose run.
* Remove the containers (`docker-compose rm`)
* Remove the images (`docker rmi image_name`) you can find the names with `docker images` 
* Build all again (`docker-compose build`)
* Run all again with the development docker-compose settings (`docker-compose up`) or with the production docker-compose settings (`docker-compose -f docker-compose.yml -f docker-compose-prod.yml up`)
* Verify that the application is accessible at http://127.0.0.1 (If your port 80 is already in use stop the application that uses it before)
* Read the INSTALL.md to learn how to run the Django development server.
* You also need to run all the application migrations and add the superuser again. 

### New translatable strings


manumilou:
@erozqba Looks interesting! Some remarks though:

* How do you specifiy which docker-compose configuration file to use when starting the containers ?
* Why using Nginx+Gunicorn for the development configuration ?
* Can you remove the po updates ? It does not belong to this PR, and make it harder to read

We could work on this together next wednesday at the Bug Squashing Party, if you can make it.

manumilou:
@cormier @madmath I ping you on this, if you'd like to review the changes @erozqba did.

madmath:
I don't have access to my laptop until tomorrow, but I'd love to test this out and review.

Sorry about the Django.po noise, I'll look into how to mitigate this.

erozqba:
@manumilou :+1:  
I will update the PR to made a rebase over dev and to correct some of your remarks.

INSTALL.md
```diff
@@ -53,14 +60,23 @@ docker exec -it [container_id] bash
...
+
+# Access to the development server
+http://0.0.0.0:8080
+
+# Access to Nginx server
+http://127.0.0.1
+
 # Run existing migrations
-python3 manage.py migrate
+python manage.py migrate
```
madmath: Could you explain in the pull request why the python3 -> python change? IMO, `python3` is clearer.

tools/nginx/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM nginx
+COPY nginx.conf /etc/nginx/nginx.conf
```
madmath: Please add newline

tools/nginx/nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+            proxy_connect_timeout 90;
+            proxy_redirect off;
+
+            proxy_set_header Host $host;
+            proxy_set_header X-Real-IP $remote_addr;
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        }
+
+    }
+}
```
madmath: please add newline

tools/mariadb/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
madmath: can we not add the mariadb directory in tools/.gitignore?

docker-compose.yml
```diff
@@ -1,16 +1,38 @@
...
+    build: tools/nginx
+    command: /bin/bash -c "nginx"
+    ports:
+      - "80:80"
+    volumes:
+      - /code/src/sous-chef/static:/usr/share/nginx/html
+    volumes_from:
+      - web
+    depends_on:
+      - web
```
madmath: please add newline

docker-compose-prod.yml
```diff
@@ -0,0 +1,34 @@
...
+    build: tools/nginx
+    command: /bin/bash -c "nginx"
+    ports:
+      - "80:80"
+    volumes:
+      - /code/src/sous-chef/static:/usr/share/nginx/html
+    volumes_from:
+      - web
+    depends_on:
+      - web
```
madmath: Please add newline

tools/mariadb/.gitignore
erozqba: yes :+1:

tools/nginx/Dockerfile
erozqba: :+1:

tools/nginx/nginx.conf
erozqba: :+1:

docker-compose.yml
erozqba: :+1:

docker-compose-prod.yml
erozqba: :+1:

INSTALL.md
erozqba: The application use python3 and in the docker container python and python3 is the same, it looks redundant for me the python3.

madmath:
Hey sorry for the review spam, I'm testing out the new review feature.

erozqba:
@madmath thanks a lot for the review, nothing to be sorry at all!

docker-compose-prod.yml
```diff
@@ -0,0 +1,34 @@
+version: '2'
```
madmath: According to the explanation for this on the [docker site](https://docs.docker.com/compose/extends/), it would make sense to have 

docker-compose.yml for the common stuff
docker-compose.override.yml for the dev-specific stuff (development server)
docker-compose.prod.yml for the prod-specific stuff

Pretty sure it would end up being less redundant, what do you think?

erozqba: Yeah, I will try to apply that.

erozqba:
Hi @madmath can you take a look again?

docker-compose-prod.yml
madmath: Small nit: I think it should be docker-compose.prod.yml according to their style guide.

INSTALL.md
```diff
@@ -40,6 +40,13 @@ $> git clone https://github.com/savoirfairelinux/sous-chef
 $> docker-compose build
 $> docker-compose up
```
madmath: I ran those two commands and when doing 'up', the database fails to initialize for me. Do you get the same error in a clean setup?

`db_1     | The latest information about mysql_install_db is available at`
`db_1     | https://mariadb.com/kb/en/installing-system-tables-mysql_install_db`
`db_1     | Initializing database`
`db_1     | mkdir: cannot create directory '/var/lib/mysql//mysql': Permission denied`
`db_1     | Fatal error Can't create database directory '/var/lib/mysql//mysql'`

It appears to be a mac specific issue as found on [Stack Overflow](http://stackoverflow.com/questions/37357315/mariadb-docker-container-on-os-x-cant-create-database-directory). I'm looking for a solution...

madmath: After installing Docker Beta for Mac, I was able to make this work. The only change was in docker-compose.yml. Are you ok with this?

```
(snip)
  web:
    restart: always
    build: .
    depends_on:
      - db

  nginx:
    restart: always
    build: tools/nginx
    command: /bin/bash -c "nginx"
    ports:
      - "80:80"
    volumes:
      - ./src/sous-chef/static:/usr/share/nginx/html
    depends_on:
      - web
```

madmath: If you are OK with this, please update the instructions to mention running [Docker for Mac beta](https://docs.docker.com/docker-for-mac/).

erozqba: Hi @madmath, I don't get that error, only Linux in my laptop, sorry.

madmath:
If my proposed change to docker-compose.yml works for you on Linux, then I can say it works for me on Mac and we should make it. See above

erozqba:
Hi @madmath can you test that change in the docker-compose.override.yml ? Because I don't think we need to mount a volume from the host in the production environment. I like more the idea of mounting the volume from the web container.
```
volumes:
  - ./src/sous-chef/static:/usr/share/nginx/html
```
Another option will be to use a separate docker-compose.mac.yml file, what option do you like more?

madmath:
Yep, this works fine! I tried to send you a pull request with the fix but I think it's messed up because of the big rebase. Anyway, if you apply the fix, this looks good to me (Y)

erozqba:
@madmath I have tested running the production version
 `docker-compose -f  docker-compose.yml -f docker-compose.prod.yml up` 
and the volume of the web container is used, there are not volume from the host used. You can test too?

madmath:
```
ERROR: for nginx  Cannot start service nginx: Mounts denied: 
The path /code/src/sous-chef/static
is not shared from OS X and is not known to Docker.
```

I don't know enough about volumes to know how to fix :(

erozqba:
@madmath https://docs.docker.com/docker-for-mac/osxfs/#/mounts

madmath:
The Dev configuration works fine. It maps the host code. But your production config which should use the container volume doesn't work (and shouldn't need to depend on osxfs). Possibly a bug in docker beta. Feel free to submit this anyway.

erozqba:
@manumilou @cormier you would like to continue testing this or is OK?

docker-compose-prod.yml
```diff
@@ -0,0 +1,8 @@
...
+services:
+  web:
+    command: /bin/bash -c "cd src && python manage.py collectstatic --noinput && /usr/local/bin/gunicorn sous-chef.wsgi:application -w 2 -b :8000"
+    volumes:
+      - /code/src/sous-chef/static
```
madmath: perhaps should just be /code/src/static

erozqba:
Hi @madmath thanks a lot for your comments, I will look at this as soon as possible.

ssirois:
@erozqba Hey there! I have tried to test this branch without success. When you have a minute, can we synch and see what's up?

I can't load the delivery initial data. Foreign key problem.

docker-compose.override.yml
```diff
@@ -0,0 +1,12 @@
+version: '2'
+services:
+  web:
+    command: python src/manage.py runserver 0.0.0.0:8000
+    volumes:
+      - .:/code
+      - /code/src/sous-chef
```
erozqba: Yes, I will move ` - /code/src/sous-chef` to the docker-compose.yml because this volume is used in production too.

tools/nginx/nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+    gzip_disable "msie6";
+
+    server {
+
+        listen 80;
+        server_name souschef.org;
+        charset utf-8;
+
+        location /static {
+            alias /code/src/sous-chef/static;
```
erozqba: :+1:

docker-compose-prod.yml
erozqba: You are right, we don't need this. :+1:

docker-compose.yml
```diff
@@ -1,16 +1,29 @@
...
       - db
+
+  nginx:
+    restart: always
+    build: tools/nginx
+    command: /bin/bash -c "nginx"
+    ports:
+      - "80:80"
+    volumes:
+      - /code/src/sous-chef/static:/usr/share/nginx/html
```
erozqba: :+1:

erozqba:
@ssirois I don't think that your issue is related to the changes made in the dokerize_app branch, we could check that today, I must go to SFL.

docker-compose.prod.yml
```diff
@@ -2,7 +2,5 @@ version: '2'
 services:
   web:
     command: /bin/bash -c "cd src && python manage.py collectstatic --noinput && /usr/local/bin/gunicorn sous-chef.wsgi:application -w 2 -b :8000"
-    volumes:
-      - /code/src/sous-chef/static
     expose:
       - "8000"
```
madmath: This "expose" clause comes up in the prod file and in the override file. Can it be brought back to docker-compose.yml?

docker-compose.override.yml
```diff
@@ -4,13 +4,7 @@ services:
     command: python src/manage.py runserver 0.0.0.0:8000
     volumes:
       - .:/code
-      - /code/src/sous-chef
-      - /code/src/sous-chef/static
     expose:
       - "8000"
     ports:
```
madmath: This exposes 0.0.0.0:8000 to outside the container, right? It's unclear to me if we need it, because we can access 0.0.0.0:80 just fine in development mode.

docker-compose.prod.yml
erozqba: Like always an excellent suggestion :+1: thanks you @madmath

docker-compose.override.yml
erozqba: You are right on this too! :+1: 
I have tested and it works just fine without the `expose: - "8000"` but I don't really understand how the **Nginx container** can redirect the request to the **Web container** through the port **"8000"** if the port is not exposed. This work just fine without the `expose: - "8000"` so I will remove that.

We will be able to access the application using "http://0.0.0.0", "http://0.0.0.0:8000", "http://127.0.0.1" or "http://127.0.0.1:8000" in development mode because the containers bind the ports `Nginx:80 -> 80/tcp`,  `Web:8000 -> 8000/tcp` to the host. 

When we use "http://127.0.0.1" and  "http://0.0.0.0" the **Nginx container** will serve the static files and redirect other requests to the **Web container** that use the **Django development server**.

When we use "http://0.0.0.0:8000" or "http://127.0.0.1:8000" the **Web container** will handle all the requests running the **Django development server**.

In production mode, we can only access using "http://0.0.0.0" or "http://127.0.0.1" because only the `Nginx:80 -> 80/tcp` port is bind and the **Web container** use **Gunicorn** instead of the **Django development server**.

erozqba:
I have applied your suggestions @madmath, you could test this again? Thanks a lot in advance! :+1:

docker-compose.override.yml
madmath: > I don't really understand how the Nginx container can redirect the request to the Web container through the port "8000" if the port is not exposed

This is because all the containers are on the same network and can see each other, so nginx can see port 8000 without any "expose" clause. The "expose" clause is to expose ports to the outside world (which you do correctly now).

madmath:
Thanks, on dev both 8000 (dev server) and 80 (nginx) work, and building the prod config only 80 works (nginx). I think this lgtm! 👍

PR #463, opened 2016-09-14
Feature/provide contact infos for client referent

### LINKED ISSUE #449: Provide contact information for referant in New Client form (clsc case worker)
### Expected Behaviour

We need to add 2 contact information entry boxes (1 for phone number and 1 for extension) for the referant in the New Client form.

### Actual Behaviour

Right now, there's only space for a name and the CLSC or Hospital making the reference.

### Steps to reproduce

none

philippemouchel:

* Provide 3 fields to fill contact informations when creating a new member as client referent

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

- When creating a new client, create a new member as referent, and fill one or more of these 3 contact fields
- You should see them when consulting client referent after creation

### New translatable strings

### Additional notes

### DIFF
--- a/src/member/templates/client/partials/forms/referent_information.html
+++ b/src/member/templates/client/partials/forms/referent_information.html
@@ -32,6 +32,29 @@
               {{form.lastname}}
           </div>
         </div>
+        <div class="field">
+          <label>Contact information</label>
+          <div class="three fields">
+            <div class="field">
+              <div class="ui left icon input">
+                <i class="at grey icon"></i>
+                {{form.email}}
+              </div>
+            </div>
+            <div class="field">
+              <div class="ui left icon input">
+                <i class="building icon"></i>
+                {{form.work_phone}}
+              </div>
+            </div>
+            <div class="field">
+              <div class="ui left icon input">
+                <i class="call icon"></i>
+                {{form.cell_phone}}
+              </div>
+            </div>
+          </div>
+        </div>
       </div>
     </div>
   </div>
--- a/src/member/templates/client/partials/forms/referent_information.html
+++ b/src/member/templates/client/partials/forms/referent_information.html
@@ -37,7 +37,7 @@
           <div class="three fields">
             <div class="field">
               <div class="ui left icon input">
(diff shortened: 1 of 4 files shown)


src/member/templates/client/partials/forms/referent_information.html
```diff
@@ -32,6 +32,29 @@ <h4 class="ui dividing header">{% trans 'Referent' %}</h4>
               {{form.lastname}}
           </div>
         </div>
+        <div class="field">
+          <label>Contact information</label>
+          <div class="three fields">
+            <div class="field">
+              <div class="ui left icon input">
+                <i class="at grey icon"></i>
```
manumilou: No need to use to the _grey_ class. Semantic UI will style the icon as part of an input entry.

src/member/tests.py
```diff
@@ -786,6 +798,8 @@ def test_form_save_data_same_members(self):
             "client_wizard-current_step": "referent_information",
             "referent_information-firstname": "Same",
             "referent_information-lastname": "User",
+            # "referent_information-email": "same@user.com",
+            # "referent_information-work_phone": "514-514-5145 #514",
             "referent_information-work_information": "CLSC",
```
manumilou: Why commenting out this ? Does it cause a problem when running the tests?

src/member/views.py
```diff
@@ -590,6 +590,33 @@ def save_referent_information(self, client, billing_member, emergency):
...
+                    member=referent,
+                )
+                ctc_work_phone.save()
+            if ref_cell_phone:
+                ctc_cell_phone = Contact.objects.create(
+                    type=CELL,
+                    value=ref_cell_phone,
+                    member=referent,
+                )
+                ctc_cell_phone.save()
```
manumilou: 1. You don't need to call save() when using create()

2. You should use the method `add_contact_information`, provided by the Member model: https://github.com/savoirfairelinux/sous-chef/blob/dev/src/member/models.py#L134.

PR #465, opened 2016-09-15
client: add notifications when clients are modified

### LINKED ISSUE #383: Implement a flash message system (success, error and warning)
### Expected Behaviour

Need a flash message system to confirm/warn/alert us about actions in app.
Here is HTML code : http://semantic-ui.com/collections/message.html#warning
We should have **success**, **error** and **warning**.
Info isn't required.

For example. a message could confirm that you have correctly edit client infos, or alert you about an error in placing an order.

### Actual Behaviour

Nothing but a way to print message per client (but it,s not a flash message).

cormier:

### DIFF
--- a/src/member/views.py
+++ b/src/member/views.py
@@ -4,12 +4,14 @@
 import csv
 from datetime import date
 import json
-from django.core.urlresolvers import reverse_lazy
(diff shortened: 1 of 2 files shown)


madmath:
Please fill the pull request template sections with a description of your change.

PR #470, opened 2016-09-16
fix455 meal labels initial version

### LINKED ISSUE #455: Generate and print meal labels
### Requirements

- Labels are for today's orders
- One label for each main dish for a client, not only for special meals
- Label contains : Client last name and abbreviated first name, Delivery date, Route name, Main dish name, Main dish size (if large), Restrictions (if any), Preparations (if any)
- Add button on Kitchen count report step page that generates PDF file and opens new browser tab to view it

lamontfr:

* delivery/views.py : generate labels file as pdf; download labels
* requirements.txt : pylabels package added (uses ReportLab)

### Status

- [X] READY (will adjust label positioning and size when I get actual sheet from SantropolRoulant)

### How to verify this change

- Run Kitchen Count as usual; then on Report page, click "Download labels"

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ django-leaflet==0.18.1
 django-extra-views==0.8.0
 transifex-client>=0.12.2
 django-template-i18n-lint>=1.2.1
+pylabels==1.2.1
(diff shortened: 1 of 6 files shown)


madmath:
Very cool. @lamontfr please document in code/in the UI which type of label sheet you support. An improvement in the future would be to support even more sheet types or label sizes.

lamontfr:
@madmath Thanks for your good suggestion. I will go to Santropol Roulant to get the actual label sheet that they use,, then adjust my layout and document the code accordingly.

PR #485, opened 2016-10-18
Add more member update steps.

### LINKED ISSUE #393: I should be able to update every client information from the update form
### Expected Behaviour

I should be able to update every information from the update form

### Actual Behaviour

Only basic information and address information is editable.

### Steps to reproduce

* Go to a client view
* Click Edit on the left column
* Try to edit other tabs than basic and address information
...

erozqba:

* Add payment_information update step.
* Add referencing update steps.
* Add emergency_contact update step.

### Status

- [X] READY

### How to verify this change

*When viewing the client information, click on "Edit" and update any information you want, then go back to see the same client and verify that the information is updated.*

### Additional notes

Remove member's duplicate template.
Add base class to client's update and view views.
Use a base class to group same code from "update client" views,
"view client" views and "update client" tests.
Remove duplicated templates to update member information, reuse the
template that exist already to add new members.

Ref: issue #393


src/member/templates/client/view/payment.html
```diff
@@ -18,15 +18,32 @@ <h2 class="ui header">
             <div class="sub header">{% trans "Billing and payment information" %}.</div>
         </div>
     </h2>
+    <div class="ui right small floated menu">
+        <a class="edit icon item" href="{% url 'member:member_update_payment_information' client_id=client.id %}">
+            <i class="icon edit"></i>{% trans "Edit" %}
+        </a>
+    </div>
```
manumilou: Rather use this markup for all the Edit buttons:

```
<h2 class="ui header">
    <div class="content">
        {% trans 'Billing' %}
        <div class="sub header">{% trans "Billing and payment information" %}.</div>
    </div>
    <a class="ui right floated button" href="{% url 'member:member_update_payment_information' client_id=client.id %}">
        <i class="edit icon"></i>
        {% trans "Edit" %}
    </a>
</h2>
```

The _float_ is set on the anchor, wrapped in the <h2>

Issue #491, opened 2016-10-24
Format phone number display


manumilou:
### Expected Behaviour

The phone numbers should all displayed in the same manner. 
The proposed format here is: XXX-XXX-XXXX.
### Actual Behaviour

No standard format is used. The phone number is displayed as entered.
For the migrated data, phone numbers are imported in the database without any whitespace: XXXXXXXXXX.
### Steps to reproduce
- Import the fixtures
- Go on the client view
- Edit the client phone number by adding whitespace (for instance)
- See how it is displayed on the client view

fpeyre:
I will make it with issue #330 (They have a form field for phone number who displayed it in the good format)

manumilou:
@fpeyre This issue is also related to the display. Besides the field validation, phone numbers must be displayed in a consistent way. For instance, if phone numbers are imported through a script with no spaces, they should be displayed properly as well. Maybe a template tag?

PR #495, opened 2016-10-28
Improve the verication of member field

### LINKED ISSUE #330: Client form: improve fields validation
### Expected Behaviour

The following fields should be properly validated:
- Postal code
- Phone number
- Email
### Actual Behaviour

No validation (only required)

fpeyre:
- Use forms.EmailField to improve email verification in the form.
- Use the library django-localflavor who give CAPhoneNumberField, CAPostalCodeField to verify the good format of a phone number and postal code number (more information here : https://django-localflavor.readthedocs.io/en/latest/localflavor/ca/ ). CAPhoneNumberField display the value with the format we desired in issue #491 
- Overload print function for phone number. CAPhoneNumberField register the number in the good format, but we can have inside the databse number with the old format. They must be print in the rigth format inside our templates
### Status
- [X] READY
### How to verify this change
- Test a postal code field. He must respect the canadian postal code synthax
- Test a phone field with and without extension (#). The phone can be wrote as XXXXXXXXXXX, XXX XXX XXXX or XXX-XXX-XXXX
-  The phone number is displayed in the same manner for each client (XXX-XXX-XXXX) with its extension if needed
- [X]  Migration is needed for this change

### Additional notes
We must rewrite the clean method of CAPhoneNumberField to include phone number with extensions
We must overload the print function of contact and member field in link to a phone number to correctly display it inside our template


manumilou:
Needs review guys!

manumilou:
Pretty good @fpeyre . 
Can you add a formatter even if the phone number is not well formatted in the database ?

src/member/models.py
```diff
@@ -151,6 +151,9 @@ def add_contact_information(self, type, value, force_update=False):
         updated. Otherwise, it should create a new one.
         """
         created = False
+        if value is None:
+            value = ''
+
         if force_update or value is not '':
             contact, created = Contact.objects.update_or_create(
```
erozqba: @fpeyre, question here: Why not use `if force_update or value is not None:` instead of adding the if condition?

fpeyre: Hmm it is to fix test who broke with error message indicate we try to insert NULL value for the field 'value' inside the database.
But sometimes we can have force_value = true with value=None, who raise the exception. So i consider None as an empty string to not raise the error.

fpeyre:
@manumilou I add a formater even if the phone number is not well formatted in the database . You can review it.

manumilou:
@fpeyre Please limit your commit message to 80 characters. I'll review your code while you are doing that :-)

PR #513, opened 2016-11-11
Remove hardcoded URL from javascript code, issue #494.

### LINKED ISSUE #494: No hardcoded URLs should be used in Javascript code
### Expected Behaviour

Every URL in the application should use Django routing system. No URL should be hardcoded, either in django templates or Javascript.
### Actual Behaviour

Some URLs are hardcoded in the Javascript code here. See this comment for reference: https://github.com/savoirfairelinux/sous-chef/pull/478#discussion_r85555738

erozqba:

* Remove hardcoded URL from javascript code

### Status

- [X] READY

### How to verify this change

* Search inside `frontend` application for hardcoded URL example `$.get('/member/client/' + $(this).val() + '/meals/preferences')` and verify that all hardcoded URL are removed.

### Additional notes

Add a custom Select class that inherits from the class form.Select
to add the data-url attribute to the options for the select #id_client
in the route : "order:create_batch"

Ref: issue #494

### DIFF
--- a/src/frontend/js/order.js
+++ b/src/frontend/js/order.js
@@ -27,7 +27,7 @@ $(function() {
     $('#id_delivery_dates').hide();
 
     $('#id_client').change(function() {
-        $.get('/member/client/' + $(this).val() + '/meals/preferences', function(data) {
+        $.get($(this).attr('data-url'), function(data) {
             if (!$.isEmptyObject(data)) {
                 $('#id_main_dish_default_quantity').val(data['maindish_q']);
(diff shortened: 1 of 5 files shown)


manumilou:
@erozqba No need to add the issue number in the commit message. Adding it in the commit body is enough and should be the only way to reference a github issue.

Can you also update _src/order/templates/list.html_ as it uses the old url _createBatch_ ? It makes the orders list page crash.

erozqba:
@manumilou done!

Issue #381, opened 2016-08-25
Gamification: allow users to select a profile picture


manumilou:
### Expected Behaviour

A system user should be able to select a profile picture: could be an upload, or a choice among a selection. But should be fun and simple :-)
Gravatar maybe ?
### Actual Behaviour

A profile image is hardcoded.
### Steps to reproduce
- Open the application
- See the image profile on the right menu

rjacquemin:
I will check this using 
http://django-avatar.readthedocs.io/en/latest/

it support gravatar

PR #515, opened 2016-11-11
#381 add django-avatar so  users can select a profile picture

rjacquemin:

*  use package django-avatar in order to handle avatar upload files
*  if nothing uploaded will try to use gravatar according user email 
* I override the default upload page to extend sous-chef base page 
* all avatar pics are uploaded in the media folder

### Status

- [x] READY

### How to verify this change
Just run the application, profile picture should be the user gravatar if nothing a default picture is displayed.
Click on the avatar picture to be redirected to the management page, you can upload a new pics and also delete old ones.  

### Additional notes

A new dependency has been added. The containers must be rebuild.


erozqba:
Tested and work as expected, +1 to improve the layout. I could help with the implementation but I'm not good making designs.

rjacquemin:
improved layout 

FYI vertical divider is buggy on chrome 
https://github.com/Semantic-Org/Semantic-UI/issues/4342

PR #519, opened 2016-11-18
Fix delivery route sheet sequencing

### LINKED ISSUE #516: In Organize Routes page, Next button produces incorrect sequence on Route Sheet
### Expected Behaviour

- In the Organize Routes page, the user chooses a route with the pull-down menu, then SousChef automatically proposes an optimal delivery sequence that the user can then modify if needed.
- When the user clicks the "Next - Delivery Route Sheet" button, SousChef displays the Delivery Route Sheet exactly in the sequence that was displayed on the Organize Routes page.
- (REVISED : NOT NEEDED) When the user clicks the "Back" button on the step 5 "Route Sheet" page, SousChef will remember the sequence for that Route and show it again in the Organize Routes page.
- SousChef will remember the current delivery sequence for each Route so that it can be retrieved on the same day if necessary from other UI pages.

### Actual Behaviour

- When the user clicks the "Next - Delivery Route Sheet" button, SousChef displays the Delivery Route Sheet in a different sequence which is not what we had on the Organize Routes page.
- When the user clicks the "Back" button on the step 5 "Route Sheet" page, SousChef displays the first route and a different sequence.

### Steps to reproduce

- Perform steps 1, 2, 3 of Kitchen Count, click "Next _ Organise Routes", choose a Route using the pull-down, change the sequence of clients on the map, Click "Next - Delivery Route Sheet".

### Proposed solution

- When user clicks the "Next - Delivery Route Sheet" button, SousChef saves the current sequence for the current Route for the day.

lamontfr:
- .gitignore : ignore pdf
- delivery/views : save chosen sequence in Organize routes to make it available for Route sheet; allow to retrieve route sequence elsewhere in UI.   (REVISED - NOT NEEDED : show address number in leaflet not just street name.)
- member/models : save only one sequence per Route.
- frontend/js/custom-leaflet : save sequence when pressing "Next".
### Status

- [X] READY

### How to verify this change

Select Kitchen Count Report in the UI and perform steps 1, 2, 3, 4, 5.


src/delivery/views.py
```diff
@@ -591,36 +671,48 @@ def dailyOrders(request):
                     'member': "{} {}".format(
                         order.client.member.firstname,
                         order.client.member.lastname),
-                    'address': order.client.member.address.street
+                    'address': "{} {}".format(
+                        order.client.member.address.number,
+                        order.client.member.address.street)
```
manumilou: Actually, number is not used anymore, and should not be used anymore.

src/frontend/js/custom-leaflet.js
```diff
@@ -32,13 +32,13 @@ function printMapAndItinerary() {
...
 
-function getWaypoints(routeId) {
+function getRouteWaypoints(routeId) {
     var waypoints = [];
     // Reset current waypoints
     control.setWaypoints(waypoints);
 
     // Ajax call to get waypoint according route
-    $.get( "../getDailyOrders/?route="+routeId, function(data ) {
+    $.get( "../getDailyOrders/?route="+routeId+"&mode=euclidean", function(data ) {
```
manumilou: How to reload a saved route on the map, if we always display the optimized route ?

lamontfr: I will call you.

lamontfr: I implemented the view to reload a saved route, but did not add this functionality in the template, since I saw last week that you were working on a new "route organizer UI" and wanted to talk to you beforehand.

PR #525, opened 2016-11-25
Change button to generate orders

manumilou:

An AJAX call is now used to generate the orders.
The AJAX is only useful to display a loading status
on the page. As the computation may take a few seconds,
it is better to alert the user that the system is working.

### Status

- [ ] READY
- [ ] HOLD
- [x] WIP (Work-In-Progress)

### How to verify this change

*Fill out this section so that a reviewer can know how to verify your change.*

### New translatable strings


src/frontend/js/delivery.js
```diff
@@ -7,4 +7,12 @@ $(function() {
         window.location.replace($url+value);
     });
 
+    $('.button.orders').click(function(){
+        $('.button.orders i').addClass('loading');
+        var url = $(this).attr('data-url');
+        $.get(url, function( data ) {
+            window.location.reload();
```
erozqba: @manumilou, I don't get this.. we make an ajax call to generate the orders to then reload the page? Why not send the orders created in the AJAX response and update the DOM instead recharging the page and make another request to the server?

src/delivery/templates/review_orders.html
```diff
@@ -116,18 +116,5 @@ <h1 class="ui header">{% trans "Review Orders" %}</h1>
...
-<script type="text/javascript">
-    $('.orders.button')
-        .popup({
-            position : 'left center',
-            target   : '.ui.statistic',
-            title    : 'Orders number for today',
-        })
-    ;
-</script>
-
```
erozqba: @manumilou thanks for moving this from here to the fronted package! :+1:

src/frontend/js/delivery.js
manumilou: @erozqba I have been lazy, I was going to implement exactly that. The rationale behind using an AJAX call is to display a visual hint to the user that the page is loading, as it may be quite some time on production's setup.
How do you propose to reload the DOM with the updated orders list ?

erozqba: @manumilou `data` should contain the list of orders generated something like:
```
data = { 
   {
      id : "978",
      client_id: "611",
      client_full_name: "Kevin Tracy",
      date: "Nov. 28, 2016",
      route: "Mile-End",
      status: "Ordered",
      price: "9.00",
      url: "/order/view/978/"
   },
   {...}
}
```
You could use jQuery to iterate over each `oder` in `data` and add it to the `tbody` element in tha page.. ?

PR #538, opened 2016-12-01
Fix #444

### LINKED ISSUE #444: Provides an option field attached to an order status change
### Expected Behaviour

I should be able to select a reason when I change the status of an order to No Charge. The reason must be mandatory and should be selected among a list of reasons.

Possible reasons:
- Restrictions mistake
- Delivery mistake
- Really unhappy client
- .....
### Actual Behaviour

The status of an order can be changed without specifying any reason.

lingxiaoyang:

* Functional: order status change tracking and reason required for no charge.
* Frontend: show a modal window to perform order status change and fill reason field.
* Backend: new model `order.OrderStatusChange`, new form, view function rewritten, new related unit tests.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Go to `http://localhost:8000/order/view/<order.pk>/` and make status changes. When the status is to be changed to `No Charge`, the reason field changes to multiple radio boxes and an `other` textarea field.
* Go to `http://localhost:8000/admin/order/orderstatuschange/` to view all OrderStatusChange objects.

Needs migration on `order` app, for the addition of new model.

### New translatable strings

Yes. Included in `order` app.

### Additional notes

For quick reference:

capture du 2016-12-01 09-58-14

capture du 2016-12-01 09-58-25
capture du 2016-12-01 09-58-29
capture du 2016-12-01 09-59-06


src/order/tests.py
```diff
@@ -432,6 +436,92 @@ def test_anonymous_user_gets_redirected_to_login_page(self):
...
+        response = self.client.post(
+            reverse('order:update_status', kwargs={'pk': self.order.id}),
+            data,
+            HTTP_X_REQUESTED_WITH='XMLHttpRequest',
+        )
+        self.assertTrue(b"errorlist" in response.content)
+        self.order.refresh_from_db()
+        self.assertEqual(self.order.status, 'B')
+
+
```
manumilou: Great job!

src/order/views.py
```diff
@@ -157,14 +157,50 @@ def get_success_url(self):
...
+        messages.add_message(
+            self.request, messages.SUCCESS,
+            _("The status has been changed")
+        )
+        return response
+
+    def get_success_url(self):
+        return reverse(
+            'order:view', kwargs={'pk': self.kwargs.get('pk')}
+        )
```
manumilou: we should add a django message somewhere to indicate the user that the order's status has been changed. Have a look at the client's status change logic.

lingxiaoyang: The lines 193-196? In fact, I copied the same code from [member/views.py](https://github.com/savoirfairelinux/sous-chef/blob/dev/src/member/views.py#L1358-L1360). Do you mean a different message?

lingxiaoyang:
It's fixed now. The earlier Ajax 200 response ate the message so it was not shown after reloading.

@manumilou Request for review.

image

PR #540, opened 2016-12-01
kitchen count: use only active clients in CM

### LINKED ISSUE #539: Only active clients should be included in the kitchen count
### Expected Behaviour

Only orders of active clients should be included in the kitchen count.

### Actual Behaviour

If the status of a client is changed after his order has been created, it will still appear in the kitchen count.

### Steps to reproduce

* Generate orders for the kitchen count
* Review the orders
* Pick one order and change the client's status to PAUSED
* Review the orders
* The order should not appear in the list

manumilou:

* Orders considered as _Shippable_ include now only active clients

### Status

- [X] READY

### How to verify this change

* See #539.

### New translatable strings

### Additional notes

Issue #539


manumilou:
@erozqba Can you review this?

src/order/tests.py
```diff
@@ -100,6 +111,18 @@ def test_order_str_includes_date(self):
...
+        Should return all shippable orders for the given date.
+        A shippable order must be created in the database, and its ORDER_STATUS
+        must be 'O' (Ordered) and the client status must be client.ACTIVE
+        """
+        client = self.orders[0].client
+        client.status = client.PAUSED
+        client.save()
+        new_orders = Order.objects.get_shippable_orders()
+        self.assertNotEqual(len(self.orders), len(new_orders))
+
```
erozqba: @manumilou, nice work! I think we could add a test to check this and avoid future regressions.  so I have added and small tests.

Issue #531, opened 2016-12-01
Revise Meal Labels Contents


lamontfr:
### Requirements (from meeting with JFF and JV on 2016-11-30)

- All labels should show : Client name, Delivery Date, Route
- If there is no ingredient clash with the main dish, show : the main dish menu name ; otherwise show : "Special _____________________" and the ingredients that clash (this will be the same combination of clashing ingredients that was shown on the Kitchen Count Report). 
- If there is an ingredient clash with the Sides, show : the client's restrictions that clash.
- If there is a meal preparation requirement : show it.
- If the meal serving is "LARGE" : show it.
- If the client has restrictions other than the clashes above : show the client's other restrictions.
- If the main dish menu name is shown according to the rules above and if there is enough room on the label, list the main dish ingredients.

### Revision 1 of requirements (from meeting with JFF and JV on 2016-12-07)

- The first line will show the Client name, the Route (centered on the line) and the Delivery date. If the route name is too long, an abridged name  will be have been added to the Database.
- The second line will show the main dish name (or '_________' if there are Main Dish clashes) and if the serving size is LARGE, this will be printed at the right end of the second line.
- The third line will start with 'SPECIAL' and will show the Dish clashes (this will be the same combination of clashing ingredients that was shown on the Kitchen Count Report - in Normal font) and the Preparations (in Bold font); if too long to fit, the text will wrap. If no Dish clashes and no Preparation, the line will be skipped and not left blank.
- The fourth line will show the Sides clashes (as specified by the client); if too long to fit, the text will wrap. If empty, the line will be skipped and not left blank.
- The fifth line will show the Other restrictions (as specified by the client); if too long to fit, the text will wrap. If empty, the line will be skipped and not left blank. "The other restrictions" = "the ingredients to avoid" + "the restricted items" - "the sides clashes".
- The main dish Ingredients will be printed only if there are no Dish clashes AND no Sides clashes.
- Il the client ordered many servings, there will be a separate label for each serving.

### Revision 1a of requirements (minor changes after PR #568, to be confirmed at next meeting with JFF and JV)

- The first line will show the Client name, the Route (centered on the line) and the Delivery date. If the route name is too long, an abridged name  will be have been added to the Database.
- The second line will show the main dish name (or '_________' if there are Main Dish clashes) and if the serving size is LARGE, this will be printed at the right end of the second line.
[...]

PR #543, opened 2016-12-02
fix529 Pick Sides Ingredients separately

### LINKED ISSUE #529: Add "Sides" selection to Kitchen Count / Pick Ingredients
### Requirements  (from meeting with JFF and JV on 2016-11-30)

- The main dish has its own ingredients which do not include the "Sides" which are two vegetables placed in separate sections of the meal container.
- SousChef must detect ingredient clashes in the Main Dish and list them on the Kitchen Count Report.
- SousChef must also detect ingredient clashes in the Sides and list them on the Labels. (see Issue #531)

### Expected Behaviour

- In the Pick Ingredients page, we should have one new multiple selection dropdown, for "Side Ingredients".
- The dropdown should allow filtering, ideally showing the ingredients whose name contains a string of characters. The latter be handled by Issue #532.

### Actual Behaviour

- All ingredients pertain to the Main dish.

lamontfr:

* delivery / fixtures / sample_data : added one component having component group = "sides" for use in tests.
* delivery / forms , delivery / templates / ingredients : new field and validation
* delivery / views : handle Sides ingredients, save them in today's menu (for clash detection to be added in the meal labels generation - this will be covered by fixing Issue #531.)
* meal / models : new component group choice = 'sides'.
* order / models : in order creation, skip component group choice = 'sides'. 
* src / meallabels.pdf : delete file such that it will not be tracked by git anymore.

### Status

- [X] READY

### How to verify this change

* In Django Admin, create a new Component named "Day s Sides" and set its component group = "Sides". (Do this only once).
* In Kitchen Count step 2, choose Sides ingredients, click "Next - Print Kitchen Count", go in Django Admin look for the Component names "Day s Sides", see its Component Ingredients for today's date.


src/delivery/forms.py
```diff
@@ -15,9 +15,24 @@ class DishIngredientsForm(forms.Form):
     )
 
     ingredients = forms.ModelMultipleChoiceField(
-        label=_('Select Ingredients:'),
+        label=_('Select Main Dish Ingredients:'),
```
manumilou: No need to use camel case here. Just: _Select main dish ingredients_ should be fine.

src/delivery/forms.py
```diff
@@ -15,9 +15,24 @@ class DishIngredientsForm(forms.Form):
...
         queryset=Ingredient.objects.order_by(Lower('name')).all(),
         widget=forms.SelectMultiple(
             attrs={'class': 'ui fluid search dropdown'}),
         required=False,
     )
+
+    sides_ingredients = forms.ModelMultipleChoiceField(
+        label=_('Select Sides Ingredients:'),
```
manumilou: No need to use camel case here. Just: _Select sides ingredients_ should be fine.

lamontfr: @manumilou I made the two changes as you requested.

PR #548, opened 2016-12-02
Use only ongoing clients when generating orders

### LINKED ISSUE #546: Orders should only be generated for ongoing clients
### Expected Behaviour

Only ongoing active clients should be considered when automatically generating the orders.

### Actual Behaviour

All active clients (both episodic and ongoing) are considered when automatically generating the orders.

### Steps to reproduce

* Change the delivery type for one client from ongoing to episodic, and make he has a delivery planned for the day
* Run `python manage.py generateorders <currentday>`
* See that on order is still being created for him

manumilou:

* Use ongoing clients rather than all active clients

### Status

- [X] READY

### How to verify this change

* See #546


erozqba:
This looks good to me!

PR #549, opened 2016-12-04
WIP This way looks more maintainable.

erozqba:
Hi @manumilou,

I have tested your PR #525  It works very well!
I look to the work of @lingxiaoyang and I get another idea to do the same but in a more maintainable way because we don't have to build HTML in javascript files.
Let me know what do you think? 

If you prefer to merge your current work, is OK, do it! I have tested it and works as expected!


src/order/models.py
```diff
@@ -154,7 +154,9 @@ def auto_create_orders(self, delivery_date, clients):
             try:
                 # If an order is already created, skip order items creation
                 # (if want to replace, must be deleted first)
-                Order.objects.get(client=client, delivery_date=delivery_date)
+                order = Order.objects.get(client=client,
+                                          delivery_date=delivery_date)
+                orders.append(order)
```
manumilou: Nice trick! This way, we can not replace completly replace the table containing the orders.

manumilou:
No, this is a nice addition @erozqba . I was annoyed by the fact to duplicate the html markup in the JS file. 
The problem of the delete button that does not work for lines added through ajax is now applcable for every orders, but it needed to be fixed anyway.

PR #550, opened 2016-12-04
cm: use standard print buttons on each step

### LINKED ISSUE #535: Display a print button on every step of the kitchen count
### Expected Behaviour

To be consistent, each step of the kitchen count that produces an artefact should provide a **Print** button.
The **Print** button should either simulate a CTRL-P (Printing capability of the browser) or trigger a PDF generation.

### Actual Behaviour

The _Route sheet_ step and the _Organize route_ view has no button to print the route sheets.

manumilou:

A print button must be a labeled button with icon.
The title attribute must be set and explain which
artefact is going to be produced.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Go through the kitchen count steps

### DIFF
--- a/src/delivery/templates/kitchen_count.html
+++ b/src/delivery/templates/kitchen_count.html
@@ -17,11 +17,17 @@
 
 
 <div class="ui basic segment no-print">
-    <a href="javascript:window.print()" class="ui labeled icon right big button"><i class="print icon"></i>{% trans 'Print the Report' %}</a>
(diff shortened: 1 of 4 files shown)


lamontfr:
@manumilou All steps work and print perfectly on my laptop. The buttons have a purple border and a printer icon at the left.

PR #554, opened 2016-12-07
Fixed #533: organize route view retrieve manual changes

### LINKED ISSUE #533: The organize route view does not display the changes made manually
### Expected Behaviour

If I change the deliveries order on the _Organize route_ view, the order should be preserved and reloaded when I reload the page, or come back later.

### Actual Behaviour

The new order is properly saved, as it appears on the route sheets, but is not displayed on the map.

### Steps to reproduce

* Go to the kitchen count
* Make sure that every order is properly geolocated, as it will affect the _organize route_ view 
* Go the step 4 and select a route with more than 2 steps
* Change the order, by drag-and-dropping the letters in front of the client's name, in the right column
* Reload the page: the order has been resetted.

lingxiaoyang:

* Add GET parameter `if_exist_then_retrieve` for getting manual route changes.
* Add this parameter in JS

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

In "organize route" in kitchen count, it should initialize with optimized route. Then do manual changes and refresh. The order of delivery should remain the same.

none

### New translatable strings

none

### Additional notes

none

### DIFF
--- a/src/delivery/tests.py
+++ b/src/delivery/tests.py
@@ -262,10 +262,16 @@ class RouteSequencingTestCase(SousChefTestMixin, TestCase):
         response = self.client.post(reverse_lazy('delivery:save_route'),
                                     json.dumps(dic),
                                     content_type="application/json")
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(b'OK' in response.content)
         response = self.client.get(
             '/delivery/getDailyOrders/?route=' +
             str(self.route_id) + '&if_exist_then_retrieve=true')
-        self.assertTrue(b'Dallaire' in response.content)
+        self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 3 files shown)


src/delivery/tests.py
```diff
@@ -264,16 +264,17 @@ def test_save_route_and_retrieve(self):
                                     content_type="application/json")
         response = self.client.get(
             '/delivery/getDailyOrders/?route=' +
-            str(self.route_id) + '&mode=retrieve')
+            str(self.route_id) + '&if_exist_then_retrieve=true')
         self.assertTrue(b'Dallaire' in response.content)
```
manumilou: This does not really assure of a proper order right? It only says that this member is in the route.

PR #563, opened 2016-12-09
Fix 560

### LINKED ISSUE #560: The add order form should be unique
### Expected Behaviour

There should be only one form to add orders, and it should allow to add every kind of order item.
The user should be able to add a _Visit_ to an order, for instance.

### Actual Behaviour

The _Add order_ form is complex and not user-friendly but allow the user to add any kind of order items.
The _Multiple orders_ form is better, by allowing to add multiple orders at one time. But it does not allow the user to add other kind of order items than meal components. 

### Steps to reproduce

* Go to _Orders_
* Click _Multiple orders_
* Try to add a visit, or a delivery to the order

lingxiaoyang:

* Expand order batch creation form: add toggles for delivery/pickup/visit
image
* On order list view, remove "new order" and rename "add multiple orders" to "add orders".
image
* Fix default component type as "meal_component" (previously "delivery").
* Fix order detail view display (replaced attribute names with display names).
image
* Add more cases in unit test concerning delivery/pickup/visit.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Create batch order -> select delivery/pickup/visit -> view orders.

### New translatable strings

- [x] If applicable, I have included updated .po files with new strings (see http://goo.gl/kfBO7N)

### Additional notes

none


src/order/templates/list.html
```diff
@@ -12,11 +12,8 @@ <h1 class="ui header">{% trans "Orders" %}</h1>
 </div>
 
 <div class="ui padded basic segment">
-    <a href="{% url 'order:create' %}" class="ui big button">
```
manumilou: Our standardized CSS classes for buttons are: 

> ui labeled icon right yellow big button

lingxiaoyang: It's fixed now in 9b3ba38.

PR #573, opened 2016-12-13
Fix #571 -- Allow printing all the route sheets at once

### LINKED ISSUE #571: Allows printing all the route sheets at once
### Expected Behaviour

An admin user should be able to print out all the route sheets from the Routes view (_/delivery/routes/_).
A button should be present at the top of the list.

### Actual Behaviour

During the kitchen count process, an admin user can print out the route sheet from the Route Sheet page. Only one route can be printed at once. 

Note: The browser printing feature is used right now.

ellmetha:

* Add a button in the "Routes Information" page allowing to print all the route sheets

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Go to the "Routes Information" page and use the print button

### New translatable strings

- [X] I have updated .po files with new strings

### DIFF
--- a/src/delivery/locale/en/LC_MESSAGES/django.po
+++ b/src/delivery/locale/en/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-11-22 23:01+0000\n"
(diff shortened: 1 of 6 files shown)


manumilou:
@lingxiaoyang  A large button would be preferable.

ellmetha:
@manumilou I've added a new version of the fix in f1c3d56.

lingxiaoyang:
Verified f1c3d56.

PR #574, opened 2016-12-13
Fix 572

### LINKED ISSUE #572: Order price: one side dish should be included in the meal's price
### Expected Behaviour

1 main dish should come with 1 free included side dish, a side dish being a green salad, a fruit salad, etc ...

### Actual Behaviour

Every side dish is billed, with no regards to the number of main dishes.

### Steps to reproduce

* Create an order
* Save
* Look at the order detail

lingxiaoyang:

* Remove "Sides" field in order batch create page: http://localhost:8000/order/create/batch (and also in unit test data)
* In `Order.objects.create_order()`, deduct free side dishes according to the number of main dishes. (1 main dish comes with 1 free side dish). The free side dishes are set `billable_flag=False`.
* Unit test.
* Add `transaction.atomic` on `create_order()`: if any error occurs inside this function, the db will revert automatically.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

1. Go to order batch create page. Create a new order with main dish and side dish.
2. Check created order. There should be non-billable side dishes.

none

### New translatable strings

none

### Additional notes

none

### DIFF
--- a/src/order/forms.py
+++ b/src/order/forms.py
@@ -8,7 +8,7 @@ from extra_views import InlineFormSet
 
 from member.models import Client
 
-from meal.models import COMPONENT_GROUP_CHOICES
+from meal.models import COMPONENT_GROUP_CHOICES, COMPONENT_GROUP_CHOICES_SIDES
 from order.models import Order_item, SIZE_CHOICES, OrderStatusChange, \
     ORDER_ITEM_TYPE_CHOICES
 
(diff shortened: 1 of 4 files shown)


lingxiaoyang:
I need to write another unit test to check when main_dish is greater than side_dishes.

lingxiaoyang:
Now it's ready.

PR #580, opened 2016-12-14
Fixed #577. Fixed #578

### LINKED ISSUE #577: Client meal default getter is too strict on JSON data.
### Expected Behaviour

The `Client.meal_default_week` JSON field should allow missing some fields, when it's zero-valued.

### Actual Behaviour

The `Client.meals_default` function is too strict on this condition: 

https://github.com/savoirfairelinux/sous-chef/blob/dev/src/member/models.py#L619

We should set a default as zero.

### Steps to reproduce

1. Create a client.
2. In django admin, remove one field like 'compote_friday_quantity'
3. Go to meal preference view page, and on Friday, it will show the system default (1 main dish + 1 dessert), regardless of the existing configuration.

### LINKED ISSUE #578: Remove "Sides" field in client meal preference configuration
### Expected Behaviour

We shouldn't configure this field.

### Actual Behaviour

We have this field.
image

### Steps to reproduce

1. Go to any client's meal preference settings (or create new)
2. Expand any days (Mon, Tue, ... Sun), there'll be a "Sides" field.

lingxiaoyang:

* Treat "sides" as a special value in meal components and remove it from client preference form.
* `Client.meals_default` now allows None fields in JSON field `Client.meal_default_week`. It sets the value as 0 in the output.
* Unit test on behaviors of `Client.meals_default` and `Client.meals_schedule`.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

1. Go to Client preference form, the "Sides" should not exist anymore.
2. Manually modify `Client.meal_default_week` in Django admin. Remove some zero fields. Then go to Client preference view, and it should not show the system default but the remaining values in this JSON field.

none

### New translatable strings

none

### Additional notes

none

### DIFF
--- a/src/member/forms.py
+++ b/src/member/forms.py
@@ -8,7 +8,7 @@ from localflavor.ca.forms import (
     CAPhoneNumberField, CAPostalCodeField
 )
 from meal.models import Ingredient, Component, COMPONENT_GROUP_CHOICES, \
-    Restricted_item
+    Restricted_item, COMPONENT_GROUP_CHOICES_SIDES
 from order.models import SIZE_CHOICES
 from member.models import (
     Member, Client, RATE_TYPE, CONTACT_TYPE_CHOICES, Option,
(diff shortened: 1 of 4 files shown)


manumilou:
The components with 0 items still should not be displayed:

capture d ecran de 2016-12-14 15-10-00

Issue #23, opened 2016-03-10
Add system of alert/notification


RignonNoel:
When a volunteer has a problem during delivery it filled an information paper for Katrina. It would be important to consider in this new version.

## Tasks

 - Add a model Notifications

## Fields

 - Description : TextField
 - Member : ForeignKey

manumilou:
Module Notification added in https://github.com/savoirfairelinux/santropol-feast/pull/33

PR #591, opened 2016-12-15
Deleting a route do not delete its clients

### LINKED ISSUE #585: Deleting a route should not delete associated clients
### Expected Behaviour

I should be able to delete a route without deleting every clients assigned to it. 
The clients should just have an unassigned route.

### Actual Behaviour

Deleting a route deletes every clients assigned to this route.

### Steps to reproduce

* Go to the Django backend
* Select the Routes model
* Try to delete a route (ex: McGill West)

fpeyre:

* Use on_delete attribute inside the models to change the default behavior when we delete a foreign key
* When the route is deleted, the attribute route of the client is set to NULL
* Add a test for this case

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Go to the Django backend
* Select the Routes model
* Delete a route (ex: McGill West)
* Verify for a client who had McGill West for route if it still exist and if it's route is null.

### New translatable strings


fpeyre:
Change done
I create a new migration (#23)

PR #600, opened 2016-12-21
Fix #601 (route=None in kitchen count)

### LINKED ISSUE #601: Client with route=None in kitchen count
### Expected Behaviour

This edge case should be considered as:

1. In `member.models`, the `route` field is always nullable and blankable.
2. It is possible to accidentally cause this edge case, either by direct edit in Django admin, or by route deletion (refer to #591).

This case should be well handled. For example, we should throw an error to warn the user at the first step (order generation), and not let it proceed to next steps as we never know what will happen.

### Actual Behaviour

"Generate orders" button is broken if there's a client with `route=None`. The error occurs during the evaluation of `route.pk` (NoneType object has no attr pk).

### Steps to reproduce

1. Go to Django admin, set route=None to an active ongoing client.
2. Go to kitchen count. Click on Generate Orders.

lingxiaoyang:

* Consider when `client.route is None`, when sorting the generated orders.
* If a client has null route, the corresponding field will show a warning sign, linking to client update form. (originally, the warning sign is shown when a client is not geolocalized.)
image

* New unit test of the display of warning signs.
* New unit test of the entire kitchen count procedure, to test if `route=None` clients are excluded in each step.
* Fix the problem that `route=None` clients are not excluded in `clashing ingredients` table.
* Also exclude all non-geolocalized clients in next steps of the kitchen count.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

1. Generate today's delivery orders.
2. Choose a client, go to Django admin, set route to null.
3. Follow the steps in kitchen count. This client should now be excluded in the procedure.
4. For ungeolocalized clients (client.member.address.latitude/longitude is None), there's a warning in organize route page, but they are included in the procedure.

none

### New translatable strings

 - [x] included

### Additional notes

none

### DIFF
--- a/src/delivery/locale/en/LC_MESSAGES/django.po
+++ b/src/delivery/locale/en/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-12-27 17:50+0000\n"
+"POT-Creation-Date: 2017-01-09 20:16+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -58,7 +58,7 @@ msgid "Restore recipe"
 msgstr ""
 
 #: delivery/templates/ingredients.html:82
-#: delivery/templates/kitchen_count.html:96 delivery/templates/routes.html:65
+#: delivery/templates/kitchen_count.html:102 delivery/templates/routes.html:65
 msgid "Back"
 msgstr ""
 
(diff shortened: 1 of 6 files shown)


lingxiaoyang:
@manumilou  I rethink it and I feel what I'm doing here may not be a good solution. We should exclude these `route=None` clients in the first step in kitchen count - the entire `delivery` package is not designed for the `route=None` case. I fear that my PR is going to cause more problems...

Maybe I close this and add an issue?

manumilou:
You have an interesting point. They actually should be excluded from the process, but, just in case, we should warn the user as well that some clients have been excluded because of no route set.
It could be a warning message, or an extra table or an highlighted line in the existing table.

lingxiaoyang:
@manumilou  updated.

manumilou:
@lingxiaoyang We should not block the whole process because a client has no route. I guess it may happen in the real life. But:

- It's important to warn the user (OK)
- To not include these clients in the rest of the process

lingxiaoyang:
@manumilou  I understand. Here's the current behavior (manually tested, I'll write new unit tests):

`route=None` clients are excluded in kitchen count components and kitchen count labels, and also in route sheets. However, they are not excluded in clashing ingredients and the restriction table. (I think this should be corrected?)

`.is_geolocalized=False` clients are included in all steps, including the final route sheets where their addresses will also be shown. However, when organizing a route, they are excluded - this is the only place that excludes ungeolocalized clients. (Is this the correct behavior?)

lingxiaoyang:
@manumilou  updated.

lingxiaoyang:
@manumilou  I updated this PR with dev branch and cleaned the commit histories.

manumilou:
@lingxiaoyang We'll sort this out with Santropol today.

manumilou:
@lingxiaoyang 

`route=None` clients should be excluded from every step of the CM process. Only the review step should highlight them.

Same thing for the `.is_geolocalized=False` clients.

lingxiaoyang:
@manumilou  Non-geolocalized clients are excluded now.

manumilou:
Sorry @lingxiaoyang , but we need a rebase :-)

lingxiaoyang:
@manumilou  Rebased in 020c4bd

PR #624, opened 2017-01-16
Fixed #614: Added categories to note.

### LINKED ISSUE #614: Define "status" of a staff note
### Expected Behaviour

A staff note should have a "status" field (values to be provided by Santropol). This field should also be editable.

### Actual Behaviour

A staff note only has "Priority" as normal or urgent. It is not possible to change the priority after its creation.

### Steps to reproduce

NA

aaboffill:

* Added `NoteCategory` model to the `Note` model.
* Note's priority field has been improved, removing CharField `priority` for a ForeignKey with a new model `NotePriority`.
* Forms, Filters and Tests has been updated.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* We can access to `Staff Notes` in the `Left Menu` and include or filter notes using the new `priority` and `category` fields.

### New translatable strings

- [x] I have included updated .po files with new strings


src/note/models.py
```diff
@@ -21,15 +21,31 @@ def get_queryset(self):
         )
 
 
-class Note (models.Model):
+class NotePriority(models.Model):
+    DEFAULT_PRIORITY_ID = 1
```
lingxiaoyang: "Normal" is not always 1!

It's better to have a lambda function as the default that returns the priority=normal object.

src/note/models.py
```diff
@@ -21,15 +21,31 @@ def get_queryset(self):
...
+    def __str__(self):
+        return u"%s" % self.name
 
-    PRIORITY_LEVEL = (
-        (PRIORITY_LEVEL_NORMAL, _('Normal')),
-        (PRIORITY_LEVEL_URGENT, _('Urgent')),
-    )
+
+class NoteCategory(models.Model):
+    DEFAULT_CATEGORY_ID = 1
```
lingxiaoyang: Also, not always 1!

src/note/migrations/0004_notepriority.py
```diff
@@ -0,0 +1,33 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def set_initial_priorities(apps, schema_editor):
+    # set NotePriority initial
+    NotePriority = apps.get_model('note', 'NotePriority')
+    NotePriority.objects.create(name="Normal")
+    NotePriority.objects.create(name="Urgent")
```
lingxiaoyang: @manumilou  Will this change affect your data migration? (note priority)

aaboffill: While this PR is not merged, `Normal` and `Urgent` are values manually created in the python code.  After this PR this values become into a Django model. I respected the original values in the migration files. 
Now, there are an important thing, we probably need a model translation solution for the nomenclatures as (`NotePriority` and `NoteCategory`)

lingxiaoyang: @aaboffill That's normally correct in our context here, but the default should be more explicit as it is not that easy to trace to the migration code to find the value.

And also - what would happen if a user deletes the "Normal" object by accident?

src/note/models.py
aaboffill: I understand your point @lingxiaoyang, but we can change now the value of that field in the `Django` admin, modifying `NotePriority` model. I was force to make that because there are some places where is necessary to add a `Note` with the `priority` and `category` fields empty. I don't complete sure what's the best option, (fix the `ID` or fix the `NAME` value field). I think that's a team! work in order to be sure that it's the perdurable decision. Whatever the decision, I can add the changes if is necessary.

lingxiaoyang: @aaboffill  My primary point is: when I read `DEFAULT_PRIORITY_ID = 1`, it is harder for me to find what it means than using, for example, a function to lazy-evaluate the default value.

Sometimes it's possible that PK doesn't start from 1, or even PK is a UUID field or a string field... but in our context it's safe. We may need to add a restriction on Django admin, to disable the "delete" button for the default values.

I have no preference on the both implementations. @manumilou Do you have a preference?

PR #637, opened 2017-01-20
Fixed #626: Restored Member's quick search inside Client's form.

### LINKED ISSUE #626: Member quick search seems doesn't work fine. Search results cached.
### Expected Behaviour

Show a list with the members that match with the introduced query in the member search field. Inside the `Client` form, in order to include a `Member` as referent.

### Actual Behaviour

The `Semantic`'s search component save in cache the first search's result, and after, the same first result will be received no matter what is the introduced query.

### Steps to reproduce

* Click on `Client` item menu.
* Select to include a new `Client` or find an existing `Client` in order to modify it.
* Go to the `Referent` tab and try to find different members using several queries in the quick `Member` search field. "No matter what is the introduced query, the result will be always the same."

aaboffill:

* Set to false the cache settings of `semantic` search. 
* Changed the position to provide data-url of Member's quick search. 

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Click on Client item menu.
* Select to include a new Client or find an existing Client in order to modify it.
* Go to the Referent tab and try to find different members using several queries in the quick Member search field. The results should match with the search query.

### DIFF
--- a/src/frontend/js/member.js
+++ b/src/frontend/js/member.js
@@ -131,10 +131,10 @@ $(function() {
         $('.ui.add.form.member').transition('scale');
     }
 
-    $search_url = $('.ui.search').attr('data-url')
+    $search_url = $('.ui.search .ui.input').attr('data-url');
     $('.ui.search').search({
         apiSettings: {
-            cache : 'local',
+            cache : false,
             url: $search_url + '?name={query}',
         },
         minCharacters : 3,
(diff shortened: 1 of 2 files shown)


lingxiaoyang:
@aaboffill  That's cool! I have a small problem here. I loaded some test data, and when I searched 'John':
image

And when I searched 'John ' (there's a space):

image

So my question: is there a limit of 7 results in the search result list, and because of it the member 2221 doesn't show in the first results?

It would be better if we can display more search results in the list, as there could be many people having the same name. If it's not easy to have a fix (as it's a semantic ui element), there's no problem either.

PR #674, opened 2017-02-24
Use a different way of calculate the age for the client.

erozqba:
Ref: issue#669

* Use a different way to calculate client age based in date.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Check if the clients ages are correctly calculated one day after and before his birthday.

### DIFF
--- a/src/member/models.py
+++ b/src/member/models.py
@@ -378,11 +378,18 @@ class ClientManager(models.Manager):
 
         today = datetime.datetime.now()
 
-        return self.filter(
+        clients = self.filter(
+            birthdate__year__lte=today.year,
             birthdate__month=today.month,
             birthdate__day__gte=today.day,
             birthdate__day__lte=today.day + 7,
         ).order_by(Extract('birthdate', 'day'))
+
+        today = datetime.date.today()
+        for client in clients:
+            client.age_to_celebrate = today.year - client.birthdate.year
+
+        return clients
 
 
 class ActiveClientManager(ClientManager):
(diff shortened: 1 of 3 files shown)


erozqba:
I have used the same issue to change the way to calculate the client age_to_celebrate because is related with the client age and we don't really need a property for that, is only used once.

erozqba:
Ok, I will add some tests to increase the coverage!

lingxiaoyang:
@erozqba  Yes, it's better to add test cases for this issue.

src/member/models.py
```diff
@@ -597,34 +604,25 @@ def age(self):
         """
         Returns integer specifying person's age in years on the current date.
 
+        To compare the Month and Day they need to be in the same year. Since
+        either the birthday can be in a leap year or today can be Feb 29 of a
+        leap year, we need to make sure to compare the days, in a leap year,
+        therefore we are comparing in the year 2000, which was a leap year.
```
lingxiaoyang: Cool solution!

src/member/models.py
```diff
@@ -378,12 +378,19 @@ def get_birthday_boys_and_girls(self):
...
+        clients = self.filter(
+            birthdate__year__lte=today.year,
             birthdate__month=today.month,
             birthdate__day__gte=today.day,
             birthdate__day__lte=today.day + 7,
         ).order_by(Extract('birthdate', 'day'))
 
+        today = datetime.date.today()
+        for client in clients:
+            client.age_to_celebrate = today.year - client.birthdate.year
```
lingxiaoyang: This is definitely better in the context of getting birthday boys and girls!

PR #681, opened 2017-03-02
Fixed #508: disable HTML5 form validation on <select> elements hidden by Semantic UI

### LINKED ISSUE #508: Error message that is at the wrong place
### Expected Behaviour
The error message should be over the field it represent

### Actual Behaviour

The error message is in the top left of the screen
bug_affichage_sous-chef

### Steps to reproduce

* create a new client
* leave the route field empty
* click next

lingxiaoyang:

* Detect the hidden `<select>` elements under `ui dropdown`. Then remove the `required` attribute to prevent the browser performing a validation that causes UI problem.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Follow the instructions on #508. The misplaced HTML5 validation popup should not happen again. Instead, the server will return a message.

none

### New translatable strings

none

### Additional notes

Ref: http://novalidate.com/


src/sous_chef/templates/base.html
```diff
@@ -70,6 +70,12 @@
     {% endif %}
     <!-- WARNING :  Do not move this line before a manual import of script! -->
     {% block extrajs %}{% endblock %}
+
+    <script type="application/javascript">
+      $(function () {
+        for(var f=document.forms,i=f.length;i--;)f[i].setAttribute("novalidate",i)
+      });  // disable HTML5 form validation
+    </script>
```
erozqba: @lingxiaoyang I think is an elegant solution to the problem, but I don't like the idea of not taking advantage of the html5 validation in the all other fields where the validation works as expected. 
Why not try to identify the fields where the html5 validation doesn't work as expected and remove the "required" attribute or the one that forces the validation?
I have tested a small modification to your change in the client creation form and I can have html5 validations in all the other fields except in the "route", for this one we can relay in the server validation.
`for(var f=document.querySelectorAll('select'),i=f.length;i--;)f[i].removeAttribute("required",i)`

lingxiaoyang: @erozqba  Because browsers implement the html5 validation differently (as for the issue, Firefox and Chrome behave differently on hidden fields). But the most important reason is that we can't configure the behavior or the stylesheet of error messages...

erozqba: @lingxiaoyang In most parts the html5 validation work just fine, the only issue is with the fields hidden by semantic-ui and replaced with a most "nice" look. I'm ok with removing the html5 validation from this hidden fields, but remove all client validation seems unnecessary to me and will affect the performance of the application, we don't need to hit the server when we already know that we have missing information or validations problems.

lingxiaoyang: @erozqba  That makes sense. I'll look for a way of working with hidden fields then...

erozqba: @lingxiaoyang Thanks :+1:

lingxiaoyang:
@erozqba  I reduced to `<select>` elements under `ui dropdown` as it seems most of the problem is causing by the behaviour of dropdown. This now bypasses the browser validation on `<select>` elements but the HTML5 validation still works on other elements.

lingxiaoyang:
Comments: the other solution would be providing a "type=hidden" attribute on Django-rendered HTML pages, which requires the following change:

From:
````python
    gender = forms.ChoiceField(
        choices=GENDER_CHOICES,
        widget=forms.Select(attrs={'class': 'ui dropdown'})
    )
````

To:
````python
    gender = forms.ChoiceField(
        choices=GENDER_CHOICES,
        widget=forms.Select(attrs={'class': 'ui dropdown', 'type': 'hidden'})
    )
````

Or write our own widget class to directly use Semantic UI's [markup](http://semantic-ui.com/modules/dropdown.html). But in my view, this is more clumsy (as there are many dropdowns throughout the code) and the code will be unclear (from a reader's view, why setting it as hidden?). 

This works because HTML5 validation doesn't work on hidden inputs. However, set the "type=hidden" attribute on JavaScript after the page has loaded didn't seem to work. That's why I simply removed the "required" attribute.

PR #690, opened 2017-03-15
Fixed #687: Add a Cancel button on the kitchen count review step

### LINKED ISSUE #687: Add a Cancel button on the kitchen count review step
### Expected Behaviour

On the first step of the kitchen count, where the orders are to be reviewed, there is no option to cancel an order directly, only edit and view.
It is recommanded to cancel an order (ie changing its status to _Cancelled_) rather than deleting it, to make sure the order won't be recreated by the scheduled tasks.

### Actual Behaviour

Only an _Edit_ and _View_ button, no _Cancel_ button. The _Cancel_ button should change the order's status to **Cancelled**.

### Steps to reproduce

* Click on the Kitchen count menu item
* Click Generate Orders button (make sure you have the fixtures loaded)

peakwinter:

* Add a Cancel Order button to the orders listed on the Kitchen Count screen, Review Orders step.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Follow Steps to reproduce as noted in #687. There, click the Cancel button. The Change Status modal (same as used on the View Order page) is displayed, allowing the user to enter a comment for their change. If the modal is submitted, the order should be successfully cancelled.

### New translatable strings

Please note that there were a lot of extraneous changes to the .po files that do not relate to the one new string I added for this fix (in `src/delivery/locale`). This might be normal but it just looked weird to me ;) If these should not be added in this pull, please let me know and I can revert the commit.

### Additional notes

none

### DIFF
--- a/src/frontend/js/delivery.js
+++ b/src/frontend/js/delivery.js
@@ -22,6 +22,39 @@ $(function() {
         });
     });
 
+    $('.ui.order.cancel.button').click(function () {
+        var self = this;
+        var modalCtntURL = $(self).attr('data-url');
+        $.get(modalCtntURL, {status:'C'}, function(data, modalCtntURL){
+            $('.ui.modal.status').html(data).modal("setting", {
+                closable: false,
+                // When approving modal, submit form
+                onApprove: function($element, modalCtntURL) {
+                    var data = $('#change-status-form').serializeArray();
+
+                    $.ajax({
+                         type: 'POST',
+                         url: $(self).attr('data-url'),
+                         data: data,
+                         success: function (xhr, ajaxOptions, thrownError) {
+                             if ( $(xhr).find('.errorlist').length > 0 ) {
+                                 $('.ui.modal.status').html(xhr);
+                             } else {
+                                 $('.ui.modal.status').modal("hide");
+                                 location.reload();
+                             }
+                         },
+                     });
+                    return false; // don't hide modal until we have the response
+                },
+                // When denying modal, restore default value for status dropdown
+                onDeny: function($element) {
+                    $('.ui.modal.status').modal("hide");
+                }
+            }).modal('setting', 'autofocus', false).modal("show");
+        });
+    });
+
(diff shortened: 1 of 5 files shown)


src/frontend/js/delivery.js
```diff
@@ -22,6 +22,38 @@ $(function() {
         });
     });
 
+    $('.ui.order.cancel.button').click(function () {
+        var modalCtntURL = $('.ui.order.cancel.button').attr('data-url');
```
lingxiaoyang: It should be `$(this).attr()`. Otherwise it'll always return the first `data-url`.

src/frontend/js/delivery.js
```diff
@@ -22,6 +22,38 @@ $(function() {
...
+        $.get(modalCtntURL, {status:'C'}, function(data, modalCtntURL){
+            $('.ui.modal.status').html(data).modal("setting", {
+                closable: false,
+                // When approving modal, submit form
+                onApprove: function($element, modalCtntURL) {
+                    var data = $('#change-status-form').serializeArray();
+
+                    $.ajax({
+                         type: 'POST',
+                         url: $('.ui.order.cancel.button').attr('data-url'),
```
lingxiaoyang: The same problem - this always returns the first `data-url`.

lingxiaoyang:
@manumilou  A question for you. Is this a good icon to use for 'cancel'?

image

manumilou:
@peakwinter You should only provide po files relevant to your changes, it is easier to read.

lingxiaoyang:
@manumilou  Sorry, I told him to do so...

@peakwinter  I was wrong about the translation files. The ones to commit are only the ones related to the module you are working on. For example, your changes concern `delivery` module, and you'll only need to commit the two po files in delivery folder. I was wrong. Sorry!

lingxiaoyang:
@manumilou  The .po file issue has been resolved and I'm merging it now...

PR #692, opened 2017-03-16
Doc: fixed dead link and added a note in i18n instruction

lingxiaoyang:


lingxiaoyang:
While waiting for manu, @peakwinter  @aaboffill  @erozqba  if two of you agree on this doc change, we may merge it.

PR #693, opened 2017-03-17
Fixed #689: Add override confirmation to "New Orders"

### LINKED ISSUE #689: Skipped command message is error-prone
### Expected Behaviour

As an administrator or a staff member, when I create orders for a given client, there is no indication that I will be creating an order for a day where an order already exists. I only get a message saying that an order has been skipped once I saved. 
The orders creation form should inform me that an order already exists for the selected days and eventually asked me if I want to override it or edit the existing one.

### Actual Behaviour

When an order is already created for a given day, the system will refuse to create an other order for this day. This is the desired behavior for the scheduled tasks to avoid order overrides. But I should be able to manually override an existing order through the order's creation form.

### Steps to reproduce

* Create an order for a client for a given day
* Use the order creation form to create an other order for the same day
* Save

peakwinter:

* Adds an override confirmation modal to the Orders > New Orders page when an order already exists for a selected day.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Go to Orders, click New orders
* Select a user name
* Select a day with an existing order
* A modal should appear informing you that an order exists for this user on the selected date
  * Selecting "Cancel" should remove the date in question from the selected dates
  * Selecting "Edit Existing" should redirect to the Update Order page for the existing order
  * Selecting "Override" should allow the user to continue, and add the date to a hidden "override dates" array. Upon completing the order process, the existing order should be deleted, and the new order created in its place.

### New translatable strings

### Additional notes

* The existing order object in the DB is deleted when the user chooses to override it.
* The user is only notified of an existing order if that order is not cancelled.
* If the order has a different status other than simply "Ordered", this status will be shown to the user in the confirmation modal.


lingxiaoyang:
@peakwinter  Is this WIP?

peakwinter:
@lingxiaoyang Only depends on the response to https://github.com/savoirfairelinux/sous-chef/issues/689#issuecomment-287180716 -- if the desired behaviour is to delete overridden orders, than this PR is ready for review

lingxiaoyang:
@peakwinter  OK, we move it to hold.

src/frontend/js/order.js
```diff
@@ -3,6 +3,35 @@ $(function() {
     // Javascript of the order application.
     // **************************************
 
+    if ($('.ui.order-override.modal')) {
+      $('.ui.order-override.modal').modal('show');
+    }
+
+    $('.order-override.cancel.button').click(function() {
+      var dateToRemove = $('.ui.order-override.modal').attr('data-order-date');
```
aaboffill: If we use an HTML data attribute `data-something="value"`, could be better to use the `jQuery` `data` method to get the HTML attribute content instead of `attr`. For example:

`$('.ui.order-override.modal').data('orderDate');`

src/frontend/js/order.js
```diff
@@ -3,6 +3,35 @@ $(function() {
...
+      if (idxOfDeliveryDate !== -1) {
+        deliveryDates.splice(idxOfDeliveryDate, 1);
+        $('#id_delivery_dates').val(deliveryDates.join('|'));
+      }
+      $('#form_create_batch #id_is_submit').val("0");
+      $('#form_create_batch').submit();
+    });
+
+    $('.order-override.override.button').click(function() {
+      var dateToOverride = $('.ui.order-override.modal').attr('data-order-date');
```
aaboffill: The same thing that above. We should use `$('...').data('something')`

src/frontend/js/order.js
```diff
@@ -3,6 +3,35 @@ $(function() {
     // Javascript of the order application.
     // **************************************
 
+    if ($('.ui.order-override.modal')) {
+      $('.ui.order-override.modal').modal('show');
```
aaboffill: The expression  `$('.ui.order-override.modal')` is used in many places, maybe is better to find the way to create the least amount instances of `$('.ui.order-override.modal')`.

src/frontend/js/order.js
```diff
@@ -3,6 +3,35 @@ $(function() {
...
+
+      if (idxOfOverrideDate === -1) {
+        overrideDates.push(dateToOverride);
+        $('#id_override_dates').val(overrideDates.join('|'));
+      }
+      $('.ui.order-override.modal').modal('hide');
+    });
+
     $('.order-delete').click(function(){
         var order_id = $(this).attr('data-order-id');
```
aaboffill: With the same idea explained above, is possible to change `.attr('data-order-id')` by `.data('orderId')`

src/frontend/js/order.js
lingxiaoyang: @aaboffill I agree, especially because this code is executed automatically after the page is ready. 

@peakwinter We may need to develop a way of managing these frontend codes, as eventually they're put and minified together. What do you suggest as the best practice on this? An app-based class/ID indicator on `<html>` element? We may also create an issue on this.

aaboffill: The idea is avoid to load the same HTML element by `jQuery` more than one time, with that, we will avoid to force to `jQuery` to go again through the XHTML. 
Some time when the HTML element is modified, we need to reload it, but if we are not in that situation, we can assign the value of $('.ui.order-override.modal') to a new `var` and use the new `var` instead of repeat the `jQuery` expression.
There are more case in this file... `$('#id_delivery_dates')`, `$('#id_override_dates')`, `$('#reason_other_field textarea')` and others...
I know that this changes does not affect the result, but we can get a more beautiful code :)

peakwinter: @aaboffill Agreed. I will make the necessary changes in the pull.

@lingxiaoyang Good idea - ultimately I think we should revisit the JS for this project from the ground up. I will create an issue for further discussion.

lingxiaoyang: @aaboffill  Ah, I misread your message.

Yes, that's true. `var $modal = $('.ui.order-override.modal')` will definitely improve the performance.

peakwinter:
@aaboffill The requested changes have been completed. b993378 has also been added per #689.

lingxiaoyang:
The modal works fine:
image

But regarding the issue text:
> The best way to handle this would be to highlight days on the calendar for which an order has already been created
Is there a way of highlighting the days that already have orders?

peakwinter:
If we want to highlight the days that already have orders, I'm blocked by #710. It's not clear to me what exactly the proper behaviour should be when the user selects a client with the dropdown, and I need to return a list of "highlighted" days from the server to process when that happens.

lingxiaoyang:
@peakwinter  I think it can be a form POST and the server returns the days to be highlighted for that client. I can do this part for you as it's me who wrote the form code.

peakwinter:
With @lingxiaoyang's latest fixes, it's my understanding that this PR is ready to go!

lingxiaoyang:
@peakwinter  Do we have to make a small change that you have noted last week?

https://github.com/savoirfairelinux/sous-chef/issues/689#issuecomment-289107538

peakwinter:
@lingxiaoyang You're right, my bad! I just pushed a commit that should resolve that comment.

lingxiaoyang:
@peakwinter  I found a bug:

I selected a client, clicked on a day that there was already an order, and I saw the Override Order prompt. I tried to click Cancel but it was always redirecting back to the Override Order prompt. Can you verify it?

lingxiaoyang:
@peakwinter  Typographical: add a space after the comma

image

lingxiaoyang:
@peakwinter  And... if possible, can we use `{% plural %}` to separate the singular and plural texts, because it would be more complicated in French?

peakwinter:
@lingxiaoyang For https://github.com/savoirfairelinux/sous-chef/pull/693#issuecomment-290425016: can you do a hard refresh? Sounds like the frontend JS might not have been updated in your environment.

For the pluralization of the template string, I couldn't find a good point in the code to get the proper internationalized plural form, so I made a compromise and used quanity + 'x'. Let me know if you think this is acceptable.

lingxiaoyang:
@peakwinter  A hard refresh did work!

As for the plural string, we can leave it as it is now. If it poses problem to a translator in the future, we can make the change then.

PR #695, opened 2017-03-17
Fixed #688: Show meal default forms for all days

### LINKED ISSUE #688: As an admin, I should be able to specify meals defaults for every clients for every day
### Expected Behaviour

In the client's preferences edit view, as an administrator or a staff member, I should be able to select meals defaults for an ongoing client even if no delivery is scheduled for a given week day.

### Actual Behaviour

For an ongoing client, as an administrator or a staff member, I can only specify meals defaults for delivery days.

### Steps to reproduce

* Edit an ongoing client's preferences
* Try to specify meals default for a non-delivery day

peakwinter:

* On the Update Client > Preferences screen, display meal default entry forms for all days, regardless if the client is marked for ongoing or episodic delivery.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Go to "Clients" and select a client.
* Click "Preferences", then next to "Meal Schedule", click "Edit".
* Change delivery type between "ongoing" and "episodic" and vice versa. The accordion menu items for all days of the week should still be displayed and editable regardless of the selected delivery type.

### New translatable strings

### Additional notes

none

### DIFF
--- a/src/frontend/js/member.js
+++ b/src/frontend/js/member.js
@@ -78,28 +78,6 @@ $(function() {
         });
     });
 
-    if($('#dietary_restriction-delivery_type select').val() == 'E') {
-        $('#form-meals-schedule').hide();
-        showAllAccordionElements();
-    } else {
-        $('#form-meals-schedule').show();
-        hideUiAccordionDays();
-        showUiAccordionSelectedDays();
-    }
-
-    $('#dietary_restriction-delivery_type .dropdown').dropdown(
-        'setting', 'onChange', function(value, text, $selectedItem) {
-            if($selectedItem.data('value') == 'E') {
-                $('#form-meals-schedule').hide();
-                showAllAccordionElements();
-            } else {
-                $('#form-meals-schedule').show();
-                hideUiAccordionDays();
-                showUiAccordionSelectedDays();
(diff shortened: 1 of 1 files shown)


lingxiaoyang:
@manumilou  Do we also need to show this information on client view, or we keep as "No delivery" as below?

image

JeanFrancoisFournier:
Meal defaults and Meal schedule are/should be 2 separate things. They should not be coupled as they prevent our "ongoing clients" from having meal defaults saved for days they don't have meal scheduled. 

This could cause clients to receive incomplete/wrong/partial deliveries when ordering out of their usual scheduled meals if CM forgets to check meal defaults for "normal" deliveries.

Does this issue takes into account the questions raised in issue 700? 
https://github.com/savoirfairelinux/sous-chef/issues/700

lingxiaoyang:
@JeanFrancoisFournier  Hello Jean-François!

Yes, the issue #700 describes this problem and the expected behaviours. At present, we are tagging it as "SRoulant_2.0". Would you need this change in the next version? If so, we can retag it as "SRoulant_1.1".

Reference: issue #688.

PR #699, opened 2017-03-20
Fixed #697: Differentiate superimposed stops and remove directions in route map

### LINKED ISSUE #697: Differentiate superimposed stops and remove directions in route map
### Expected Behaviour
- When more than one client are at the same stop (for example many addresses are in the same building) such that the alphabetic letter labels are superimposed on the map, find a way to indicate this on the map or on the sequence list that accompanies the map.
- The label text A=santropol should be changed to A=Santropol Roulant.
- There should be an extra stop at the end of the sequence such that the delivery volunteer sees the return path from the last client in the route back to the Santropol Roulant.
- Remove the "Directions" in the lower right section of the web page.
- Remove the "Directions" on the printed map.

### Steps to reproduce
- Make sure that the data contains, for a given geolocated address, more than one client (the only difference being the apartment number)
- Run the Kitchen Count steps 1, 2, 3 and 4
- Click "Organize" for a non empty route

peakwinter:

On both the route planner and printout...

* ...multiple "waypoints" (clients) at the same address show on the map as a special marker with a star in it, indicating that more than one client can be found at the same location.
* ...when an address has multiple clients, the address is only listed once, and the clients' names are comma-separated.
* ...a final waypoint has been added, showing the itinerary for the return trip to Santropol Roulant.
* ...the turn-by-turn directions have been removed.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Ensure that you have at least two clients in a single route that share a street address and a delivery on the date in question.
* Proceed through the Kitchen Count process until you arrive at the Routes Information screen.
* Browse to the route that has the two clients in it, and click Organize.
* Verify that:
  * ... one map marker with a white star appears above the street address that has two clients
  * ... the waypoint list in the sidebar contains only one entry for said address, and both clients' names show up as comma-separated.
  * ... the turn-by-turn directions do not appear
* Click "Map and Directions" and repeat the aforementioned verifications.

### New translatable strings

### Additional notes

* The mechanism to group map markers (and waypoints) by address is done by comparing the latitude and longitude of the geocoded points for the clients in question. If the latitudes and longitudes are equal to within five decimal points of tolerance, the clients will be considered to live at the same street address.
* Apartment numbers are not listed on the Delivery Route printout, rather only on the Routes Information printout. I did not change this as it was existing behaviour and the issue did not request a change in this regard.

### DIFF
--- a/src/frontend/js/custom-leaflet.js
+++ b/src/frontend/js/custom-leaflet.js
@@ -69,10 +69,13 @@ function getRouteWaypoints(routeId) {
              waypoints[i].name = data.waypoints[i].member ;
         }
 
-        //add fisrt waypoint for santropol
+        //add first waypoint for santropol
         var santro = new deliveryPoints(L.latLng(45.516564,  -73.575145));
-        santro.name = "santropol";
+        santro.name = "Santropol Roulant";
         waypoints.splice(0, 0, santro);
+
+        // add return waypoint to go back to santropol
+        waypoints.push(santro);
 
         // Set waypoints on the map
         control.setWaypoints(waypoints);
--- a/src/frontend/js/custom-leaflet.js
+++ b/src/frontend/js/custom-leaflet.js
@@ -58,15 +58,20 @@ function getRouteWaypoints(routeId) {
     // Ajax call to get waypoint according route
     $.get( "../getDailyOrders/?route="+routeId+"&mode=euclidean&if_exist_then_retrieve=true", function(data ) {
         var deliveryPoints = L.Routing.Waypoint.extend({ member:"", address:""});
+
+        var groupedWaypoints = groupWaypoints(data.waypoints);
+
         // create an array of waypoint from ajax call
-        for(var i in data.waypoints)
-        {
-             waypoints.push(new deliveryPoints(L.latLng(data.waypoints[i].latitude, data.waypoints[i].longitude)) );
-             waypoints[i].options.address = data.waypoints[i].address;
-             waypoints[i].options.member = data.waypoints[i].member;
-             waypoints[i].options.id = data.waypoints[i].id;
-             waypoints[i].name = data.waypoints[i].member ;
-        }
+        Object.keys(groupedWaypoints).forEach(function(k, i) {
+          var currentWaypoint = groupedWaypoints[k][0];
+          var memberString = groupedWaypoints[k].map(function(x) { return x.member; }).join(', ');
+          waypoints.push(new deliveryPoints(L.latLng(currentWaypoint.latitude, currentWaypoint.longitude)));
+          waypoints[i].options.multipleClients = groupedWaypoints[k].length !== 1;
+          waypoints[i].options.address = currentWaypoint.address;
+          waypoints[i].options.member = memberString;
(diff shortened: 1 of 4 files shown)


peakwinter:
This PR is ready for review.

lingxiaoyang:
Behaviors (copied from issue):

- [ ] When more than one client are at the same stop (for example many addresses are in the same building) such that the alphabetic letter labels are superimposed on the map, find a way to indicate this on the map or on the sequence list that accompanies the map.
This is marked with a "star" (@lamontfr is this ok?) and the popup shows all people and their address.
image
And in the print sheet, apartment # is not listed. @lamontfr  Is this the expected behavior?
image
- [ ] The label text A=santropol should be changed to A=Santropol Roulant.
In the list this is changed. But in the popup it says "Santro". @peakwinter 
image
- [x] There should be an extra stop at the end of the sequence such that the delivery volunteer sees the return path from the last client in the route back to the Santropol Roulant.
Verified.
- [x] Remove the "Directions" in the lower right section of the web page.
Verified.
- [x] Remove the "Directions" on the printed map.
Verified,

src/frontend/js/custom-leaflet.js
```diff
@@ -183,7 +222,7 @@ function main_map_init (map, options) {
                 var marker;
 
                 // adjust marker according waypoints
-                if (wp.name == "santropol") {
+                if (wp.santro && !wp.hideMarker) {
                     // add awesome marker for santropol
                     marker =  L.marker([45.516564, -73.575145], {
                         draggable: false,
```
lingxiaoyang: A question here (not mentioned in the issue but I think is better to do this): can we also set other markers with "draggable:false"?

peakwinter: This was explicitly added in the file in several locations, but I am also not sure if it is really necessary. In disabling draggable markers one can still create a custom itinerary by dragging the lines between markers.

lingxiaoyang: @peakwinter Really? We then keep it unchanged unless required.

lamontfr:
@peakwinter @lingxiaoyang
Re: **Question 1**: Is a star is OK as a label when there are more than one client at the same stop ?
I think that a two-character label would be preferable for example "B*" because : 
- On a route we could have more than one stop for which there are many clients;
- The delivery persons really need the map to find their way around and so in the map screenshot above they would not find stop B on the map.

Re: **Question 2**: There is no apartment number in the "Route sequence" below the map. Is this OK ?
I think that this is correct, since : 
- Many clients may live in the same building;
- The detailed route sheet (printed separately from the map) includes the client's full address with Apartment number and telephone number, which are essential.

I will meet Jean-François tomorrow and check the above with him.

lamontfr:
@peakwinter @lingxiaoyang I just finished a meeting with Jean-François and he confirmed my answers to your questions. So we are good to go.

peakwinter:
Thanks @lamontfr !

This is therefore ready to be merged if the final review passes :) @lingxiaoyang

lingxiaoyang:
@lamontfr @peakwinter  Thank you!

PR #711, opened 2017-03-23
Fixed #685: Added functionality to remove client status

### LINKED ISSUE #685: Scheduled status modifications can not be deleted
### Expected Behaviour

As a system administrator or staff member, I should be able to delete a scheduled status change for a client, from the client's view (Ex: http://localhost:8000/member/view/<client_id>/status)

### Actual Behaviour

As a system administrator or staff member, I cannot delete a scheduled status change for a client, from the client's view (Ex: http://localhost:8000/member/view/<client_id>/status)

### Steps to reproduce

* Go a client's view
* Schedule a status change by using the status button, at the top-right
* Try to delete the future status change

aaboffill:

* Changed `ForeignKey` relationship between `ClientScheduledStatus` class to `OneToOneField`, and to handle this relations as a `pair` 
* Implemented a functionality to remove the client scheduled status. 
* Included and fixes tests.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Go to the `Client` details and select the link `Status`.
* If the Scheduled Status list is empty, we can add a new Scheduled Status in the top right of the page.
* At the last column of the Scheduled Status list, we can see the button `Remove`

### New translatable strings

- [X] If applicable, I have included updated .po files with new strings (see https://github.com/savoirfairelinux/sous-chef/blob/dev/docs/internationalization.adoc)


src/member/signals/handlers.py
```diff
@@ -0,0 +1,10 @@
+from django.db.models.signals import post_delete
+from django.dispatch import receiver
+
+from ..models import ClientScheduledStatus
+
+
+@receiver(post_delete, sender=ClientScheduledStatus)
+def ensure_pair_remove(sender, instance, **kwargs):
+    if instance.get_pair:
+        ClientScheduledStatus.objects.get(pk=instance.get_pair.pk).delete()
```
lingxiaoyang: Discussion: if a pair status change has been performed, for example, yesterday, do we delete it anyways or keep it in the database?

A related question to discuss: do we allow the user to delete a status change that has happened?

lingxiaoyang: @manumilou Refs #685

src/member/apps.py
```diff
@@ -3,3 +3,9 @@
 
 class MemberConfig(AppConfig):
     name = 'member'
+
+    def ready(self):
+        # import signal handlers
+        # (every models imported inside handlers will be instantiated as
+        # soon as the registry is fully populated.)
+        import member.signals.handlers  # noqa
```
lingxiaoyang: This import looks a little bit magic as what we really want here is to make the connection of the signal and the handler. The `@receiver` decorator is not very clear on its side effect.

I suggest writing `post_delete.connect(ensure_pair_remove, sender=ClientScheduledStatus)` in the ready method to improve the readability.

aaboffill: @lingxiaoyang I understand your point, but now we have only one signal receiver, and that the receiver method have just two code's lines. But in the future maybe we can have many signal receivers with many code's lines. Maybe is inconvenient to keep all signal logic handled into the `ready` method.
I followed the suggestion explained in the 3rd cooment... http://stackoverflow.com/questions/2719038/where-should-signal-handlers-live-in-a-django-project 

what do you think?

src/member/signals/handlers.py
aaboffill: @lingxiaoyang and @manumilou, Ling has been made an interesting questions.

I think that is better avoid to remove the status if that status was processed, we can keep the status as `Operation Status: Processed`, because is totally true, and avoid to show the button `Remove` for the processed statuses.

src/member/apps.py
lingxiaoyang: @aaboffill  You can keep the function in `receivers.py`, and simply refer to that function in `ready` method.

The fact is I didn't understand this import at the first reading until noticing the side effect of the `@receiver` decorator that registers the function to Django's system. Semantically, calling `post_delete.connect()` says more clearly that we may have state changes on `post_delete` object.

aaboffill: Let me understand @lingxiaoyang,  your point is change `@receiver(post_delete, sender=ClientScheduledStatus)` by `post_delete.connect(sender=ClientScheduledStatus)` ??

aaboffill: I think that is necessary to add the `dispatch_uid`, i'll append that.

lingxiaoyang: @aaboffill In `apps.py`:

````python
from member.signals.handlers import ensure_pair_remove

...
    def ready(self):
        ClientScheduledStatus = self.get_model('member.ClientScheduledStatus')
        post_delete.connect(ensure_pair_remove, sender=CSS, dispatch_uid="...")
````

And in member.signal.handler:

````python
def ensure_pair_remove(sender, instance, **kwargs):
    if instance.get_pair:
        ClientScheduledStatus.objects.get(pk=instance.get_pair.pk).delete()
````

So split the definition of handler function and the registration of signal handler.

Comment (aaboffill) [inline src/member/apps.py]: @lingxiaoyang your proposal could be accepted, we know that both options works good. But I think that we need to separate responsibilities. Maybe the `ready()` method does not have to know what are the necessary signals to be connected, all logic or all responsibility related to the receivers should be declared inside the signal python package.

Whatever is the final decision, I think that is good to make the corresponding example into [Django model project](https://gitlab.savoirfairelinux.com/django/modele).

Comment (lingxiaoyang) [inline src/member/apps.py]: @aaboffill  We'll have your version! I'll merge it.

**Comment by lingxiaoyang** [comment]
@aaboffill I found a bug: when I tried to change a client's status in the modal window, I had an error:

````python
web_1    | [30/Mar/2017 16:13:43] "POST /member/client/100/status/scheduler HTTP/1.1" 500 14495
web_1    | Internal Server Error: /member/client/100/status/scheduler
web_1    | Traceback (most recent call last):
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/exception.py", line 42, in inner
web_1    |     response = get_response(request)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 187, in _get_response
web_1    |     response = self.process_exception_by_middleware(e, request)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 185, in _get_response
web_1    |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/views/generic/base.py", line 68, in view
web_1    |     return self.dispatch(request, *args, **kwargs)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/utils/decorators.py", line 67, in _wrapper
web_1    |     return bound_func(*args, **kwargs)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapped_view
web_1    |     return view_func(request, *args, **kwargs)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/utils/decorators.py", line 63, in bound_func
web_1    |     return func.__get__(self, type(self))(*args2, **kwargs2)
web_1    |   File "/code/src/member/views.py", line 1420, in dispatch
web_1    |     return super(ClientStatusScheduler, self).dispatch(*args, **kwargs)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/contrib/auth/mixins.py", line 56, in dispatch
web_1    |     return super(LoginRequiredMixin, self).dispatch(request, *args, **kwargs)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/contrib/auth/mixins.py", line 92, in dispatch
web_1    |     return super(PermissionRequiredMixin, self).dispatch(request, *args, **kwargs)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/views/generic/base.py", line 88, in dispatch
web_1    |     return handler(request, *args, **kwargs)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/views/generic/edit.py", line 217, in post
web_1    |     return super(BaseCreateView, self).post(request, *args, **kwargs)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/views/generic/edit.py", line 183, in post
web_1    |     return self.form_valid(form)
web_1    |   File "/code/src/member/views.py", line 1439, in form_valid
web_1    |     response = super(ClientStatusScheduler, self).form_valid(form)
web_1    |   File "/code/src/order/mixins.py", line 26, in form_valid
web_1    |     response = super(FormValidAjaxableResponseMixin, self).form_valid(form)
web_1    |   File "/usr/local/lib/python3.5/site-packages/django/views/generic/edit.py", line 162, in form_valid
web_1    |     self.object = form.save()
web_1    |   File "/code/src/member/forms.py", line 511, in save
web_1    |     self.object.process()
web_1    | AttributeError: 'ClientScheduledStatusForm' object has no attribute 'object'
````
image

Do you have the same error?

File: src/member/tests.py [hunk 3]
```diff
@@ -2024,6 +2024,42 @@ def test_view_client_status_update_no_end_date(self):
...
+            password='test1234'
+        )
+        self.client.login(username=admin.username, password='test1234')
+        data = {
+            'client': self.active_client.id,
+            'status_from': self.active_client.status,
+            'status_to': Client.STOPCONTACT,
+            'reason': 'Holidays',
+            # set today to be processed immediately
+            'change_date': date.today().isoformat(),
```
lingxiaoyang: Cool!

PR #715, opened 2017-03-30
Maintenance: update Python package dependencies

lingxiaoyang:

Fix the dependency report on [versioneye](https://www.versioneye.com/user/projects/58814a4eb194d4003d528a43?child=58814b93b194d40038f47439#tab-dependencies).

#### Update `django-extra-views`: 0.8.0 -> 0.9.0

[Change log](https://github.com/AndrewIngram/django-extra-views/blob/master/CHANGELOG.rst)

Summary: simply added support for Django 1.10.

Where we use it:
````
src/order/views.py:from extra_views import CreateWithInlinesView, UpdateWithInlinesView
src/order/forms.py:from extra_views import InlineFormSet
````

Influences: nothing.

#### Update `django-leaflet`: 0.19.0 -> 0.21.0

[Change log](https://github.com/makinacorpus/django-leaflet/blob/master/CHANGES)

Summary: bugfixes.

Where we use it: `src/frontend/custom_leaflet.js`

Influences: nothing. We don't use the function `leaflet.draw` in an explicit manner.

#### Update `reportlab`: 3.3.0 -> 3.4.0

[Change log](https://bitbucket.org/rptlab/reportlab/src/fd9ca874bf300f810a69ce7679999e034cfc3f35/CHANGES.md?at=default&fileviewer=file-view-default)

Summary: a lot of bugfixes and new features.

Where we use it: the generation of labels PDF during the kitchen count.

Influences: the changes seem to be most internal and most of them are bugfixes. I tested the label PDF generation with reportlab 3.4.0 and it seemed fine. We don't have any geometric shapes so most of the changes can be ignored for us.
image

#### Update `gunicorn`: 19.6.0 -> 19.7.1

[Change log](http://docs.gunicorn.org/en/stable/2017-news.html)

Summary: a lot of bugfixes. No breaking changes for us.

Where we use it: starting the production server of sous-chef.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Please verify these changelogs and my analysis above.

none

Rebuild your docker container.

### New translatable strings

none

### Additional notes

none


erozqba:
:+1:

Issue #716, opened 2017-03-30
Billing creation throws an error when the month is not provided.


lingxiaoyang:
### Expected Behaviour

The user should be able to preview a month's billing, and create the billing for that month.

### Actual Behaviour

There are two problems:

~~~1. The billing preview doesn't contain any orders, and neither does the created billing. That is to say, on page http://localhost:8000/billing/add/, after selecting a month, click on "1. Preview", no orders will be displayed; click on "2. Create Billing", we're creating an empty billing of 0.00$.~~~ It seems I got this wrong. Billing is only created for delivered orders.

2. The button "2. Create Billing" and the blue button "Create Billing" at the bottom of page should not display when the month is not provided. At present, if we leave the month as blank and click on these two buttons, we will have an internal server error. We need to avoid this.

### Steps to reproduce

* step 1: Go to http://localhost:8000/billing/add/
* ~~step 2: Select a month that has orders.~~
* ~~step 3: Click "1. Preview", you'll see an empty list.~~
* ~~step 4: Click "2. Create Billing", you'll see an empty billing.~~
* step 5: Go back to step 1. Now leave the month as blank.
* step 6: Click "2. Create Billing" or the "Create Billing" button at the bottom of the page, you'll see an error.

![image](https://cloud.githubusercontent.com/assets/8630726/24523350/c562da30-1560-11e7-81ab-2c2cc5dc830c.png)

PR #718, opened 2017-03-31
Hotfix.

erozqba:

* Fix issue #716  
* ...
* ...

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

*Fill out this section so that a reviewer can know how to verify your change.*

### New translatable strings

### Additional notes

This is not the proper way to fix this, is only a temporal patch.
We should refactor all this logic to send the data using POST and not GET,
add form validations and more.

Ref: issue 716

### DIFF
--- a/src/billing/templates/create.html
+++ b/src/billing/templates/create.html
@@ -1,5 +1,0 @@
-{% extends "base_billing.html" %}
-{% load i18n %}
-{% load static %}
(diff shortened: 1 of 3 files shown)


lingxiaoyang:
Thank you!

PR #592, opened 2016-12-16
Fix #570 -- Add basic permission definition

ellmetha:
### Changes proposed in this pull request:

* Add some basic rule definition (can read data / can edit data) and use them in all views

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Create a staff user (using the `User.is_staff` field) and browse the dashboard.

### New translatable strings

- [X] I have updated .po files with new strings

PR #721, opened 2017-04-04
Fixed #720: Restricted button to remove scheduled status to admin users

### LINKED ISSUE #720: Restrict the button to remove an Schedule Status Change to admin users only
### Expected Behaviour

The `button` to remove an `Schedule Status Change` should be acceded by admin users only. 

### Actual Behaviour

There are not a condition to show the button to remove an `Schedule Status Change`.

### Steps to reproduce

* Login into the system with a not admin user.
* Go to a client's view.
* Schedule a status change by using the status button, at the top-right.
* Go to the list and the button to remove `Schedule Status Change` will be displayed.

### Notes

Related to #685 , this issue was fixed but the restriction to show the button to remove is missing.

aaboffill:

* Added restriction to show the button to remove `Scheduled Status Change` to the admin users only.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Login into the system with a not admin user.
* Go to a client's view.
* Schedule a status change by using the status button, at the top-right.
* Go to the list and the button to remove `Schedule Status Change` won't be displayed.

### DIFF
--- a/src/member/views.py
+++ b/src/member/views.py
@@ -1450,14 +1450,10 @@ class ClientStatusScheduler(
         )
 
 
-class ClientStatusSchedulerDeleteView(LoginRequiredMixin, generic.DeleteView):
+class ClientStatusSchedulerDeleteView(PermissionRequiredMixin, generic.DeleteView):
     model = ClientScheduledStatus
+    permission_required = 'sous_chef.edit'
     template_name = "client/view/delete_status_confirmation.html"
-
-    def post(self, request, *args, **kwargs):
-        if not request.user.is_staff:
-            return HttpResponseBadRequest()
-        return super().post(request, *args, **kwargs)
 
     def get_success_url(self):
         return reverse(
--- a/src/member/views.py
+++ b/src/member/views.py
@@ -1450,7 +1450,8 @@ class ClientStatusScheduler(
         )
 
 
-class ClientStatusSchedulerDeleteView(PermissionRequiredMixin, generic.DeleteView):
+class ClientStatusSchedulerDeleteView(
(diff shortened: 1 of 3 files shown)


lingxiaoyang:
@aaboffill  I'm sorry, I forgot to tell you that we have been using "rules" for the permission control: see this change in PR #592.

Code references:
https://github.com/savoirfairelinux/sous-chef/blob/dev/src/sous_chef/rules.py
In template: https://github.com/savoirfairelinux/sous-chef/blob/dev/src/member/templates/client/list.html#L15
In view function: https://github.com/savoirfairelinux/sous-chef/blob/dev/src/member/views.py#L51

Apart from this, all seems good to me. Can you change ".is_staff" to "sous_chef.edit"?

aaboffill:
@lingxiaoyang I added your suggestions using the defined rules.

src/member/views.py
```diff
@@ -1449,8 +1450,10 @@ def get_success_url(self):
         )
 
 
-class ClientStatusSchedulerDeleteView(generic.DeleteView):
+class ClientStatusSchedulerDeleteView(
+        PermissionRequiredMixin, generic.DeleteView):
```
aaboffill: @lingxiaoyang This is an example of the ugly python code when we have 79 characters as the line limit. It's my opinion!

lingxiaoyang: I think it may be unavoidable in some cases, even with a higher limit. But 79 is not enough for a Django application, for sure.

lingxiaoyang:
Verified:
image
vs
image

PR #722, opened 2017-04-04
Fixed #704: Change Member.address from ForeignKey to OneToOneField.

### LINKED ISSUE #704: Multiple clients may not share the same address instance
### Expected Behaviour

In sample_data.json, find Paul Taylor and Kevin Tracy:
https://github.com/savoirfairelinux/sous-chef/blob/dev/src/delivery/fixtures/sample_data.json#L24-L34
https://github.com/savoirfairelinux/sous-chef/blob/dev/src/delivery/fixtures/sample_data.json#L81-L91

You'll see that they have the same address instance ID 511. When this happens, there should be some sort of indication on the UI.

### Actual Behaviour

From the Sous-Chef UI, we can't know if Paul and Kevin share the same address instance. And when we modify Paul's address, Kevin's address will also be changed - again, no messages for this behavior. This would cause undesirable results.

What should we expect if this happens? Or we should use OneToOneField to ensure that clients don't share address instances? Please comment.

### Steps to reproduce

* step 1: backup database, and wipe it: `python manage.py flush`
* step 2: load sample data `python manage.py createsuperuser; python manage.py loaddata sample_data`
* step 3: change Paul's address. And observe Kevin's address.

lingxiaoyang:

* Model change: ForeignKey -> OneToOneField
* Model migration: deals with shared address instances.
* Unit test on model migration.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Before applying the migration, in Django backend admin, make two or more clients linking to the same Address instance.

Do the migration. Then check the address instances of these clients, which should then be separated.

### New translatable strings

none

### Additional notes

none


erozqba:
:+1:

PR #725, opened 2017-04-11
Fixed #691: new "Route" menu for viewing and editing delivery sequences

### LINKED ISSUE #691: Determine the sequence of all clients in a route even when they have no delivery
### Expected Behaviour (changed on 2017-03-23 - see comment)
- For a route, allow the CM to optimize, modify and save the sequence of all (active, pending or paused) clients in the route, whether they have a forthcoming delivery or not. This becomes the "default order" for the route. This will require a new main menu entry, because it is not related to the daily kitchen count. However, the same UI and most of the logic of "Route / Organize" can be reused.
- When a new client is created or a client is resumed from a "stop: contact/no contact" status, ~~the CM has to positioned him(her) in his(her) route sequence~~ the client is positioned to the end of the list by default.
- On a day where there are deliveries on this route, the stops with meals will be ordered as per the default (except for the gaps for clients having no delivery on that day), and the CM will still be able to alter the sequence temporarily for the current day if necessary.

### Actual Behaviour

- The CM can optimize and make changes to the sequence of clients having meals to be delivered in a route for a given day.
- On a another day after that, when the CM displays the sequence for the latter route, only clients that have a meal everyday are still in the correct position in the sequence.
- This means that when a client has no meal delivered on a given day, the system forgets the client's proper position and the next day on which he has a meal, he will appear at the end of the sequence. This will oblige the CM to move again the client back to his place in the sequence, which is time consuming.

lingxiaoyang:

* New model `member.DeliveryHistory`: stores the specific vehicle/client sequence/comments of routes on each delivery day.
* New main menu "Routes": provides list/detail/edit view of a `member.Route`, notably allows the editing of default delivery sequence.
* Logic change in kitchen count -> organize route: works on `member.DeliveryHistory` instead of `member.Route`. The operation "organize route" creates the `member.DeliveryHistory`. The route sheet is only available after having today's `DeliveryHistory`.
* Frontend: refactored leaflet map code, concerning (a) client creation/edit form, (b) route view/edit, (c) delivery route edit.
* Basic unit tests.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

New URL: try updating and viewing the routes.
http://localhost:8000/member/routes/

Verify the leaflet map in client creation/edit form works well.

Process kitchen count, in last step, try organising routes. After that, go back to "Route" menu and you will find delivery histories.

### New translatable strings

- [x] If applicable, I have included updated .po files with new strings (see https://github.com/savoirfairelinux/sous-chef/blob/dev/docs/internationalization.adoc)

### Additional notes

none


lamontfr:
@lingxiaoyang Your described changes meet the requirements, thank you.
However, I pulled your PR to test it. After I start the docker webserver, I run gulp in tools/gulp to rebuild the static files in sous-chef/src/sous_chef/static and I see that are generated there. The web pages work but I cannot get the map (leaflet) to show up. When I switch back to the dev branch, the map shows up. What am I doing wrong ?

lamontfr:
@lingxiaoyang Your described changes meet the requirements, thank you.
I pulled your PR to try it. I had to run gulp to rebuild the static files and to clear the browser cache to get the map to work. I will now continue testing.

lingxiaoyang:
@lamontfr  Thank you! Please keep me informed of any bugs that I may leave.

src/sous_chef/locale/fr/LC_MESSAGES/django.po
```diff
@@ -8,7 +8,11 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
+<<<<<<< 04044df6a72fe7b52a056b48de3aebdd276f9f4a
```
davidbgk: Remaining merge conflict.

lingxiaoyang: @davidbgk  Thank you! I missed it.

lingxiaoyang: Fixed in f539bab.

lingxiaoyang:
@lamontfr  Before merging, please wait until I increase the test coverage.

lamontfr:
@lingxiaoyang I am still testing. I found a minor problem.

- Go to  http://localhost:8000/delivery/routes/  the first time for a given day
- The organise buttons are all grey which is OK
- click organise on a route which has orders so we can see the clients and the map
- click cancel at the bottom
- we are back in http://localhost:8000/delivery/routes/ and the organise button is now green even though we did not click save
- in the admin, we can see that a Delivery-History instance has been created, but that its client-id-sequence is an empty list
- if we click the Route-sheet button (on the line where we have the Green organise, we notice that the clients are NOT in the sequence that we saw in the Organise list and map.
- **ANALYSIS** : The organise button should only turn green after we have clicked save for each route.
- **Proposed solution**  in file delivery/views.py , class RoutesInformation, method get_context_data , line 271 , add an if as such : 
```pyrhon
if delivery_history.client_id_sequence != []:
    has_organised = True
```

lingxiaoyang:
@lamontfr  I verified this:

> we notice that the clients are NOT in the sequence that we saw in the Organise list and map

Thank you! I'll follow your proposed solution.

lingxiaoyang:
@lamontfr  Fixed this in 32ab5db (this is a force push). I further compared the `client_id_sequence` and the delivery clients, and added a 3rd state of "invalid" when the two sets are not identical:

https://github.com/savoirfairelinux/sous-chef/pull/725/commits/cb73f5839374dd8661de435cf656ff947fe9e491#diff-2ef7e7af631d376f710a6970ea0f4b0bR276

I changed the UI here a little bit. For the three states:
1. Not organised: as before, a gray button "Organise" and a disabled "Route Sheet" button.
2. Invalid: this is the added state. It'll be red with a warning icon. The "Route Sheet" button is disabled.
3. Organised: this is changed to gray color (in my last commit it was green but it doesn't look good with a red button and the general look and feel). The user can still be reminded by the text "Organised" (before it was "Organise"), the icon and the enabled "Route Sheet" button.

image

Please let me know your feedback!

lingxiaoyang:
The unit tests are done.

lamontfr:
@lingxiaoyang Thanks, I will test further.

lamontfr:
@lingxiaoyang I tested and also showed this to Jean-François a few weeks ago. My first comment is about the terms we show the user regarding the status of each route in http://localhost:8000/delivery/routes/ . JF was questioning why we should organise each route separately and whether it could be automatic. I thought about it and I believe that it is a good thing that the users look at the day's route sequence on the map especially for the first few months of operation. This is a good verification before printing the route sheet. In the future (for milestone 2 or later) we may add a mechanism to organise all routes according to the default sequence, but not now.

What I suggest is to simply adjust the terms and the tooltips as follows : 

- The label "Organise" would be renamed "Confirm", and the tooltip "Organise the route" would be renamed "View, adjust and confirm the day's route sequence".

- The label "Organised" and the tooltip "View or edit the organised route" remain the same.

- The label "Invalid" would be renamed "Modified" (on a red button as it is now) and the tooltip "View and edit the organised route" would be renamed "Clients have changed, please reconfirm".

lamontfr:
@lingxiaoyang You did a great job, my tests were successful, except for one. The problem I found is when a change is made to a client in a route after it was "organised". In this case the day's delivery history is inconsistent with the delivery that will take place.

Here is a scenario which may happen in practice :
- At some time in the morning, the user has all his orders for the day and he starts the kitchen count.
- He gets to the route list http://localhost:8000/delivery/routes/ and organises the routes one by one.
- He finds a route A that is too busy and decides to move client Y to route B by editing the client's route assignment.
- As soon as he goes back to the route list, route A and route B will show "invalid".
- When we click on invalid to edit route B we see that client Y was added and has status "X". When we save the route, it becomes "organised".
- When we click on invalid to edit route A we see that client Y is still there with its order "DELETED" but when we save, it remains invalid. Therefore we cannot print this route sheet.

Here is what I propose :
- In member/views.py:get_clients_on_delivery_history() , when we find that the client is missing (because his order was cancelled or because his route assignment was changed), we could simply exclude him from the delivery_history. This will allow the delivery history for the route to correspond to the real deliveries of the day.
- If we want to advise the user about the removed clients in the route, the func_add_warning_message() could be used to provide a message that could be appended in a list by the caller delivery/views.py:EditDeliveryOfToday:get_context_data() and the list included in the context data of the view to be shown on the page using some kind of popup. This could wait for milestone 2.

lamontfr:
@lingxiaoyang I just found another problem. I will describe it and propose a solution tomorrow.

Please wait for my next post before working on my previous comments of today.

lingxiaoyang:
@lamontfr  OK, thank you!

lamontfr:
@lingxiaoyang The other problem I found is that when we first organise a route for today's delivery, the clients are not in the default sequence for the route. I thought that I had tested that it was working in the past, but it is not working anymore. I think that the cause is in member/views.py:get_clients_on_delivery_history() as the "non configured" clients are not sorted according to the default sequence.

Here is the change that I propose, which will also solve the problem of the "invalid" that cannot be cleared in today's route as explained in my previous comment above. Note that the behavior is preserved for displaying the past delivery history detail (when called from member/views.py:DeliveryHistoryDetailView).


member/views.py:get_clients_on_delivery_history() L1625 before : 
```python
    # First, check the clients on `id_sequence` and see if we put them in the
    # display sequence.
    for client_pk in (delivery_history.client_id_sequence or []):
        if client_pk in clients_dict:
            client = clients_dict[client_pk]
            client.has_been_configured = True
            clients_on_delivery_history.append(client)
            del clients_dict[client_pk]
        else:
            # This client has been deleted, or the related order has been
            # deleted after that day.
            # If this happens, we may let user know, because a DeliveryHistory
            # shouldn't be edited after its delivery date.
            try:
                client = Client.objects.get(pk=int(client_pk))
                client.order_of_the_day = None
                client.has_been_configured = True
                clients_on_delivery_history.append(client)
            except Client.DoesNotExist:
                if func_add_warning_message:
                    func_add_warning_message(
                        _("The delivery history of this route contains a "
                          "client (ID %(id)s). This client has been deleted "
                          "from the database and thus is ignored in the "
                          "delivery sequence.") % {
                              'id': client_pk
                          }
                    )
            except (ValueError, TypeError):
                # During `int()`
                pass

    # Then, check the clients that had delivery on that day but were not
    # configured on the delivery sequence. Append them to the display sequence.
    for client_pk, client in clients_dict.items():
        client.has_been_configured = False
        clients_on_delivery_history.append(client)
    return clients_on_delivery_history
```
member/views.py:get_clients_on_delivery_history() L1625 after : 
```python
    # First, check the clients on `id_sequence` and see if we put them in the
    # display sequence.
    for client_pk in (delivery_history.client_id_sequence or []):
        if client_pk in clients_dict:
            client = clients_dict[client_pk]
            client.has_been_configured = True
            clients_on_delivery_history.append(client)
            del clients_dict[client_pk]
        elif func_add_warning_message:
            # This client has been deleted, or the related order has been
            # deleted after that day.
            # If this happens, we may let user know, because a DeliveryHistory
            # shouldn't be edited after its delivery date.
            try:
                client = Client.objects.get(pk=int(client_pk))
                client.order_of_the_day = None
                client.has_been_configured = True
                clients_on_delivery_history.append(client)
            except Client.DoesNotExist:
                func_add_warning_message(
                    _("The delivery history of this route contains a "
                      "client (ID %(id)s). This client has been deleted "
                      "from the database and thus is ignored in the "
                      "delivery sequence.") % {
                          'id': client_pk
                      }
                )
            except (ValueError, TypeError):
                # During `int()`
                pass

    # Then, check the clients that had delivery on that day but were not
    # configured on the delivery sequence. Append them to the display sequence.
    # We use the preferred ordering for all potential clients on the route.
    route = Route.objects.get(pk=delivery_history.route.pk)
    for client_pk in route.client_id_sequence:
        client = clients_dict.get(client_pk)
        if client:
            client.has_been_configured = False
            clients_on_delivery_history.append(client)
    return clients_on_delivery_history
```
Please let me know what you think.

lingxiaoyang:
> When we click on invalid to edit route A we see that client Y is still there with its order "DELETED" but when we save, it remains invalid. Therefore we cannot print this route sheet.

I rethought about it and I agree with you, that we can simply ignore the problematic data (order cancelled, order deleted, client deleted, client changed to another route, etc.). This part is really out of our control regarding data integrity. We may simply prompt there's a problem, let user go to admin if he wants, and if the client exists, provide the link to the client file.

> when we first organise a route for today's delivery, the clients are not in the default sequence for the route

I think I made it wrong while amending the PR. The correct order is (1) consulting DeliveryHistory (2) consulting Route default (3) putting in random order, but somehow I removed (2). I'll fix it.

lingxiaoyang:
@lamontfr  Updated in a24de2b:

1. Fixed the DeliveryHistory order: added the missing Route default and fixed the related unit test.
2. Changed UI (invalid -> modified, etc..)
3. Changed the error messages in `member/views.py:get_clients_on_delivery_history()`
4. Fixed the problem in Client address form that the map is not initialized with latitude and longitude.

lamontfr:
@lingxiaoyang I tested your latest version and it works perfectly. Your solution for giving information messages to the user is very elegant. I confirm that you can now merge this PR into dev.

lingxiaoyang:
@lamontfr  Great, thank you!

PR #732, opened 2017-05-01
Update dependencies

erozqba:

* Django upgrade from 1.10.7	to 1.11
* django-leaflet upgrade from 0.21.0 to 0.22.0
* semantic-ui-calendar upgrade from 0.0.7 to 0.0.8
* Remove one line from travis.yml that make the test run twice. 

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

You will need to stop, destroy and recreate images, containers, volumes, and networks.

### New translatable strings

### Additional notes

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-django>=1.10,<1.10.99
+django>=1.11,<1.11.99
 mysqlclient>=1.3,<1.3.99
 gunicorn>=19.7,<19.7.99
 factory_boy>=2.8,<2.8.99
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -10,7 +10,6 @@ install:
 
 script:
   - PYTHONWARNINGS=once coverage run --omit=*.virtualenvs*,*virtualenv* src/manage.py test --settings=sous_chef.settings_test member meal order notification delivery note billing page
-  - python3 src/manage.py test --settings=sous_chef.settings_test_fr member meal order notification delivery note billing page
```
lingxiaoyang: I intentionally added this to test the application in a French setting. I think there were some differences in string comparison. Why do we delete this line?

erozqba: @lingxiaoyang sorry! I don't have pay attention to this "french" setting, it seems to be a double testing for me when I was looking into Travis, I will add the line again!

PR #736, opened 2017-05-05
Fixed #734: don't generate orders when it's out of client's delivery schedule

### LINKED ISSUE #734: Orders should not be automatically generated when it is out of an ongoing client's delivery schedule
### Expected Behaviour

@lamontfr Please confirm if this is an anomaly of sous-chef.

There are orders automatically generated on every midnight for all ongoing clients. The same logic is executed when the admin clicks on "Generate orders" in kitchen count review step 1. For every ongoing client, if he/she has delivery on that day, there should be an order for what he/she wants.

### Actual Behaviour

An ongoing client may select the days that he/she wants delivery as the delivery schedule, for example, Mon, Wed and Fri. Currently, sous-chef creates the order regardless of delivery schedule. For this example client, sous-chef will create orders on every day, while it should only create when it's Mon, Wed or Fri.

### Steps to reproduce

* step 1: create a test client or modify an existing client, set "delivery type" as Ongoing, select some days other than today in "schedule". Then verify if you have correct information:
image

* step 2: make sure that there's no order in "Kitchen Count". Delete them if they already exist.
* step 3: click "Generate orders" and you'll see this test client's order. It should not appear because the client doesn't have delivery today.

### Code reference

1. When clicking "Generate orders":
https://github.com/savoirfairelinux/sous-chef/blob/395ac8252a1de669d46b518f459ba1093a9d47cb/src/delivery/views.py#L1282
https://github.com/savoirfairelinux/sous-chef/blob/395ac8252a1de669d46b518f459ba1093a9d47cb/src/order/models.py#L171-L176 -- no checking for delivery schedule. In fact, it should consult `client.meals_schedule` instead of `client.meals_default`. 

2. Automatic generation at midnight:
https://github.com/savoirfairelinux/sous-chef/blob/395ac8252a1de669d46b518f459ba1093a9d47cb/src/order/management/commands/generateorders.py#L39  -- then goes to the same code.

lingxiaoyang:

* Two unit tests for midnight order generation and kitchen count refresh order, respectively.
* `Client.set_meals_schedule()` is renamed as `Client.set_simple_meals_schedule()` with code optimised.
* Fixes `Order.auto_create_orders()`: use `Client.meals_schedule` instead of `Client.meals_default`.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

See #734 

none

### New translatable strings

none

### Additional notes

none

### DIFF
--- a/src/datamigration/management/commands/importclients.py
+++ b/src/datamigration/management/commands/importclients.py
@@ -99,4 +99,4 @@ class Command(BaseCommand):
                 client.member.save()
 
                 # Add Client option meals_schedule
-                # client.set_meals_schedule([])
+                # client.set_simple_meals_schedule([])
--- a/src/datamigration/management/commands/importmeals.py
+++ b/src/datamigration/management/commands/importmeals.py
@@ -91,7 +91,7 @@ class Command(BaseCommand):
                             prefs['diabetic_' + row[day] + '_quantity'] = \
                                 int(row[11 + (delivery_day * 10) + 4])
 
-                    client.set_meals_schedule(meals_schedule)
+                    client.set_simple_meals_schedule(meals_schedule)
                     client.meal_default_week = prefs
                     client.save()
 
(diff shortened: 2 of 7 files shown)


lamontfr:
@lingxiaoyang I will check it tonight.

lingxiaoyang:
@lamontfr Thank you for review!

PR #745, opened 2017-05-17
Fixed #742: Increase the font size for multiple clients at one address.

### LINKED ISSUE #742: Use a larger symbol on map when multiple clients at one stop
### Expected Behaviour
(As requested by and discussed with Jean-François on 2017-05-12)
- On a route map, we should clearly see in the blue "drop" label when there are more than one client at a stop.

### Actual Behaviour
- The label at a stop shows the number with a relatively small asterisk compared to the size of the number. This makes it difficult to distinguish from the stops having only one client.

### Suggestions to solve issue
- Increase the font size for only the asterisk OR
- Instead of one asterisk, put four asterisks to make it stand out OR
- Change the color of the blue "drop" when there are many clients.

lingxiaoyang:

* For multiple clients' marker, increase the text font by 1.5x.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Please check:
1. Route detail view.
2. Route edit view.
3. Kitchen count - delivery route view.

none

### New translatable strings

none

### Additional notes

none

### DIFF
--- a/src/frontend/js/custom-leaflet.js
+++ b/src/frontend/js/custom-leaflet.js
@@ -168,9 +168,16 @@ function sous_chef_leaflet_map_init (map, options, settings) {
 
                 }
                 if (wps.length > 1) {
-                    // Tweak icon content
+                    // Tweak icon content and text size when there are multiple clients
+                    var MAGNIFY_FACTOR = 1.5
                     if ((wps[0].settings.icon || {}).options.glyph) {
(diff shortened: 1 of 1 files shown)


lamontfr:
@lingxiaoyang I tested all three cases and it works perfectly. You can merge this.

PR #751, opened 2017-05-25
Route sheets report as PDF

### LINKED ISSUE #696: Improve printed route sheets contents and formatting
### Expected Behaviour
- Print all the routes in one operation (one click).
- Each route should start on a odd numbered new page (see issue #611 ).
- The information block on a client stop on the route sheet should never be split by a page break.
- The font size should be large enough (Helvetica 10 or above) to be clearly legible when printing.
- The SRoulant address and contact information is required in the page layout template.
- PDF file generation would be desirable (see also comment on March 11th in #611).

(This is based on requirements given by Jean-François on 2017-03-17)

### Actual Behaviour
- The layout of the global routes printout http://localhost:8000/delivery/routes/?print=yes is too basic and is not usable by the delivery volunteers therefore the CM has to print each route separately.
- Using the print route sheet button in http://localhost:8000/delivery/route_sheet/1/, printing font size is too small.
- Page breaks on odd numbered pages are not supported by the browsers.

lamontfr:
- delivery/views.py :
  - class RoutesInformation : generate and download pdf route sheets, create context only when displaying list of delivery routes, removed route sheet rendering as html for printing.
  - defineStyles() : refactored all reportlab paragraph styles.
  - class MultiRouteReport : generate pdf containing multiple route sheets.
  - drs_make_lines() : removed date because it was not used, ensured translation of component_group name.
- order/models.py : get_delivery_list() : ensured translation of component_group name.
- delivery/tests.py : adjusted tests for route sheets pdf printing to use attributes in response header instead of context, added test for multiple pages route sheets.
- delivery/fixtures/sample_data.json : added long delivery notes to test multiple pages route sheets.
- delivery/templates/routes_print.html : removed file because we now create a pdf instead.
- delivery/templates/routes.html : changed icon and tooltip to "download".

### Status
- [X] READY

### How to verify this change
- The goal is to verify the expected behavior of issue #696.
- Make sure that a route has enough orders today and long delivery notes such that it will require more than one page to print.
- Make sure that a route needs an odd number of pages to print such that a blank page will be inserted on the back side of the last sheet.
- Click on Kitchen count
- Generate orders for the day
- Click on step 4-Routes
- Click confirm and then save for each of the routes; the button label will be "organized"
- Click on the Route Sheets button; this will download a pdf file
- Open the pdf file and verify the contents and formatting, footer text and page skips.


lingxiaoyang:
Great work, thank you!

PR #757, opened 2017-06-23
Use nodejs 8.

kousu:
## Fixes #756 

which is that gulp breaks due to half-broken dependencies which seem to have forgotten how to work under old node versions now.

### Changes proposed in this pull request:

* Switch nodejs to the 8.x series (the latest) over the 6.x series

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

*On a fresh checkout* (do something like "git clone https://github.com/kousu/sous-chef && mv sous-chef sous-chef-npm-patch && cd sous-chef-npm-patch" because `docker-compose` uses the name of the folder its in to name its containers), follow INSTALL.md before applying the patch, note that the `gulp` step fails, and then follow it after.

### Deployment notes and migration

- [ ] Migration is needed for this change

### Additional notes

See #756.

Issue #761, opened 2017-06-23
Promiscuous port config


kousu:
The port config syntax in docker-compose.yml [defaults to](https://docs.docker.com/compose/compose-file/#ports) opening on 0.0.0.0, i.e. any host can contact it directly, which is really really really silly considering how people think of docker as a protective environment.

I suggest this patch:
```
  ports:
-   - "8000:8000"
+   - "127.0.0.1:8000:8000"
```

then you can stick a proper nginx with SSL in front.

lingxiaoyang:
Agreed!

erozqba:
We should not have this type of configuration in the canonical docker-compose.yml configuration file. We only use this port binding for development, so we should add it in the docker-compose.override.yml file. For the development environment, I think that security is not a priority, we are using the development web server anyway, so we could use "8000:8000".

Issue #758, opened 2017-06-23
CI server doesn't test docker


kousu:
The INSTALL.md instructions only explain how to install using Docker. I actually found it was easier to use my knowledge of webapps to figure out how to set it up without Docker, which [as I discovered](https://github.com/savoirfairelinux/sous-chef/issues/756#issuecomment-310600425), is what the Travis-CI server thinks too; it doesn't actually test any of the Docker code, which means changes to Dockerfile or docker-compose.yml are not exercised, and more importantly, external version changes in our dependencies, when run on Docker, are not.

I would prefer if non-Docker was the install method, but if you prefer to stick with Docker then please make sure both `docker-compose up` and `docker-compose -f docker-compose.yml -f docker-compose.prod.yml` are tested.

lingxiaoyang:
@kousu I think a CI test on docker is not so critical. I agree with you that `docker-compose.override.yml` is a bad idea and it's better to make it explicit. But then we may not really need the CI test because the docker-compose file is usually straightforward and self-explaining. It would be sufficient after the PR #762. How do you think?

kousu:
I think that all the Docker steps, whether scripted or in manual instructions, are just as much part of the codebase as Django. Not being able to reliably get docker working means delaying deployments and bringing new people on board. I don't know what's involved in setting it up under Travis but I imagine it being tricky and painful; but if that isn't done, I would want a clean build (e.g. ideally in a fresh VM) to be done manually at least once a month.

erozqba:
@kousu The name convention for share docker-compose configurations files use `docker-compose.override.yml` for the default "dev" environements, check here: https://docs.docker.com/compose/extends/

lingxiaoyang:
@kousu I got it wrong. Thanks @erozqba !

erozqba:
@kousu Could you check the PR #775 and give me other suggestions that you find we should include in the Travis-CI tests?

PR #762, opened 2017-06-23
Better automate deploy

kousu:

* Automate the frontend compilation steps inside Dockerfile
* Make it so that `docker-compose up` is enough to get the project running (admittedly with an empty database.)
* Tidy up the subsidiary docker-compose files by coalescing their common bits directly into Dockerfile or the main docker-compose.yml
* Don't use docker-compose.override.yml, because explicit is better than implicit: instead name it docker-compose.dev.yml
* Add copious comments to the Dockerfile.

### Status

This was branched off #757 so please merge that first.

- [] READY
- [x] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Do a fresh checkout and a fresh build. See that it works with just `docker-compose up`. Try then with `docker-compose -f docker-compose.yml -f docker-compose.yml up`

### Additional notes

Deployment has been a huge headache for me. INSTALL.md is pretty nicely fleshed out, but it's still really hard to knowing when and what order to run things, and I would rather move as many instructions into (commented) scripts instead of into instructions that humans need to follow manually.


kousu:
^ this was a change to the Docker stuff so per #758 the CI check is misleading here. The only way to verify this is to do a fresh build on your own machine.

kousu:
Ah, actually, this isn't quite working. It puts the static files in `src/static/` instead of `src/sous_chef/static`

kousu:
I'm probably not going to get to fixing this up this week. It shouldn't be tricky, but it just needs some repetition until it reliably puts all the files in the right paths, and docker rebuilds are really really slow and discouraging. So, this will happen, but you'll have to be patient with me. Thanks.

.gitignore
```diff
@@ -106,3 +106,8 @@ src/sous_chef/static
 
 # Ignore generated pdf files.
 src/*.pdf
+
+# the magic docker-compose override file is for overriding settings locally
+# it should not be a part of the repo. If you want per-environment settings,
+# use an explicit `docker-compose -f docker-compose.yml -f docker-compose.environment.yml`
+docker-compose.override.yml
```
erozqba: @kousu If we are using docker-compose we should follow their conventions, and by default docker-compose command will read the configurations files "docker-compose.yml" and "docker-compose.override.yml".  The main idea is to have in the "docker-compose.yml" the configurations that remain the same in all environements, use "docker-compose.override.yml" to the development environment that is used more frequently by developers and have separated configurations files for other environments like staging or prod.
For more detailed information check https://docs.docker.com/compose/extends/#example-use-case

Dockerfile
```diff
@@ -2,13 +2,35 @@ FROM python:3.5
...
+RUN python3 src/manage.py collectstatic --noinput
+WORKDIR /code/tools/gulp
+RUN npm install  # gulp dependencies
+RUN gulp
+WORKDIR /code/
+
+# the default entrypoint for the container
+# when this ends, the container ends, then restarts
+# can be overridden in docker-compose.yml files
+CMD python3 src/manage.py runserver 0.0.0.0:8000
```
erozqba: We should not use this server by default, this is a development server that should be used only for development if we are going to use this image in production, we should use by default gunicorn and override it with the development server in the docker-compose.override.yml configuration file, or not have any entry point at all.

Dockerfile
```diff
@@ -2,13 +2,35 @@ FROM python:3.5
...
+# install javascript dependencies
 RUN npm install gulp -g
-ADD . /code/
-RUN apt-get clean
+
+# compile front-end "assets" (this should be unnecessary since they're basically just files, but hey, it's 2017)
+# the 'rebuild node-sass' is a patch over a glitch which showed up recently;
+# the tip that fixed it is from https://github.com/sass/node-sass/issues/1387#issuecomment-185451183
+# perhaps it is not truly necessary.
+run mkdir /code/src/sous_chef/static/
```
erozqba: If this is not truly necessary we should not include it, so we need to check it.

docker-compose.dev.yml
```diff
@@ -0,0 +1,10 @@
+# use as `docker-compose -f docker-compose.yml -f docker-compose.dev.yml up`
+# this causes the the local folder to be loop-mounted into the container
+# so that you can edit code with your normal editors and have the changes
+# immediately be reflected in the running container, without a `docker-compose build`.
+
+version: '3'
+services:
+  web:
+    volumes:
+      - .:/code
```
erozqba: We don't need this file and we should keep using docker-compose.override.yml because for developers is easier to do `docker-compose up -d` than `docker-compose -f docker-compose.yml -f docker-compose.dev.yml`

docker-compose.override.yml
```diff
@@ -1,8 +0,0 @@
-version: '3'
```
erozqba: Keep this file.

docker-compose.yml
```diff
@@ -14,6 +14,8 @@ services:
       - backend
 
   web:
+    ports:
+      - "127.0.0.1:8000:8000"
     environment:
```
erozqba: I don't think we need this binding why you have added it?

lingxiaoyang:
I think we need to reconcile this PR with #775, as both are working on the same files.

Dockerfile
kousu: I thought you must have *an* entry point. Without one, the container shuts down immediately.
I would love it if docker let you just bring up a container with no master process. It lets you `docker exec` arbitary side commands in a single container, so why do any of them need to be key?

docker-compose.yml
kousu: I moved this binding from `docker-compose.override.yml`, and added the fix for #761 at the same time.

.gitignore
kousu: Thanks for the link. Having the laser focus there really helps move the debate forward, and I agree that you're doing it just as they suggest.

I still think it's wrong. `docker-compose.yml` can have dev settings in it by default which are then overridden by `docker-compose.prod.yml`. "override" says to me "these are settings which are custom to this specific instance, which I need in order"; if they meant "override" to always mean "dev" then they should have called it "dev". Emphatically I think that name means they must not be stored as part of the codebase, and not be in the repo, be `.gitignore`'d.

lingxiaoyang: @kousu  I agree with you personally. I don't like the use of the word "override" of docker-compose either and it brought me some troubles... but I think this is their convention according to their doc, so I would follow the usage of "override" so that we work on the same page with docker-experienced developers. 

I think it is important to mention the doc link somewhere in the project, in order to help developers like me and @kousu who don't know this convention of docker-compose a priori.

Dockerfile
lingxiaoyang: I think this is resolved in #775.

Dockerfile
lingxiaoyang: Resolved in #775: 

https://github.com/savoirfairelinux/sous-chef/pull/775/files#diff-3254677a7917c6c01f55212f86c57fbfR28
https://github.com/savoirfairelinux/sous-chef/pull/775/files#diff-f772bd5403017e9075de6e9455c26521R10

INSTALL.md
```diff
@@ -58,29 +71,26 @@ $> python3 manage.py createsuperuser
...
+$> docker-compose exec web sh -c "cd tools/gulp && gulp"
 ```
 
-Please rest assured that the "unsafe-perm" option will not bring any security risk to sous-chef. The Node.js packages that we are installing here are only used for generating static files, such as images, CSS, JavaScript, etc., and will never be executed from external.
-
-If you have an error with this command, try deleting the folder `tools/gulp/node_modules` completely and rerun it. If the problem still exists, please let us know.
+If you have an error with this command, try deleting the folder `tools/gulp/node_modules` completely and rerun it. If the problem still exists, please let us know, including the versions of node, npm, and gulp that you have.
 
+[collectstatic](https://docs.djangoproject.com/en/1.11/ref/contrib/staticfiles/#collectstatic) is similar, but built into Django.
+We are not using it to manage our own static files, but some of our Django dependencies might need this to be run.
```
lingxiaoyang: @kousu  Good work! We keep this file.

docker-compose.dev.yml
lingxiaoyang: Resolved in #775.

docker-compose.override.yml
lingxiaoyang: Resolved in #775.

docker-compose.yml
lingxiaoyang: Seems to me that it's needed? I didn't see in #775.

erozqba: @lingxiaoyang if we are only using this binding in development mode, in the `docker-compose.override.yml` we don't relly need to take care of security and promiscuous ports because we are using the development server anyway.

lingxiaoyang: @erozqba  Got your idea.

erozqba:
@kousu could you rebase this branch?

PR #781, opened 2017-07-17
Maintenance: update dependency version.

lingxiaoyang:

* Update django-avatar from 3.1 to 4.0 -- the backward incompatible change has been fixed. See its [changelog](https://github.com/grantmcconnaughey/django-avatar/blob/master/CHANGELOG.rst).
* Update django-localflavor from 1.4 to 1.5. No backward incompatible changes.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Please be sure to rebuild your 'web' container.

N/A

### New translatable strings

N/A

### Additional notes

N/A


erozqba:
@lingxiaoyang could you test this with and without production settings? 
I need to make some config changes in order to make it work:

```
diff --git a/src/sous_chef/settings.py b/src/sous_chef/settings.py
index 04d6c07..4516a1a 100644
--- a/src/sous_chef/settings.py
+++ b/src/sous_chef/settings.py
@@ -202,7 +202,7 @@ FORMAT_MODULE_PATH = (
 )
 
 # Static files (CSS, JavaScript, Images)
-# https://docs.djangoproject.com/en/1.9/howto/static-files/
+# https://docs.djangoproject.com/en/1.11/howto/static-files/
 STATIC_ROOT = os.path.join(BASE_DIR, 'static')
 
 STATICFILES_DIRS = (
@@ -211,9 +211,17 @@ STATICFILES_DIRS = (
 STATIC_URL = '/static/'
 
 # Avatar files
-MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
-MEDIA_URL = '/media/'
-AVATAR_STORAGE_DIR = 'avatars/'
+if DEBUG:
+    # When using the development server, serve files directly from /media/
+    # https://docs.djangoproject.com/en/1.11/howto/static-files/#serving-files-uploaded-by-a-user-during-development
+    MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
+    MEDIA_URL = '/media/'
+else:
+    # In non development mode, serve files from /static/ using gninx as
+    # dedicated server
+    # https://docs.djangoproject.com/en/1.11/howto/static-files/deployment/#serving-static-files-from-a-dedicated-server
+    MEDIA_ROOT = os.path.join(STATIC_ROOT, 'media')
+    MEDIA_URL = '/static/media/'
 AVATAR_PROVIDERS = (
     'avatar.providers.PrimaryAvatarProvider',
     'avatar.providers.GravatarAvatarProvider'
```

The main reason is that in production settings, we are using gunicorn server for python files and nginx is serving the static files, I propose to have uploaded files like images for avatars inside 'static' because we already share a volume with the static files between the web and the nginx containers.

lingxiaoyang:
Replaced by #801 . Closing.

PR #784, opened 2017-07-26
Use nginx container only with in prod settings.

erozqba:

* Use nginx container only with in prod settings, because we don't really need for development.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* docker-compose down
* docker-compose up -d
* docker-compose ps 

The last command should only show two containers running, web and mariadb.

### New translatable strings

### Additional notes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,7 +102,7 @@ tools/gulp
 tools/mariadb
 
 # Ignore generated assets.
-src/sous_chef/static
+src/sous_chef/assets
 
 # Ignore generated pdf files.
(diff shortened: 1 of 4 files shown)


erozqba:
@lamontfr you could ask Nick to also verify this? Thanks in advance!

erozqba:
After 21 days whitout any review, I will accept this myself because this is only docker-compose configuration and not a python/django related code. If something get broken I will fix it.

PR #799, opened 2017-08-16
Allow blank mid and rid for member

### LINKED ISSUE #796: Django Admin must accept a member with blank Mid and Rid fields
### Expected Behaviour
- When we need to edit the data for a member using the admin, fields Mid and Rid should not be required.

### Actual Behaviour
- Fields Mid and Rid were used only for data migration from the Feast database. For new members created in SousChef, they are required which is wrong.

lamontfr:
* src/member/models.py : allow blank values

### Status

- [X] READY

### How to verify this change
- Admin / Members / select a member X / check that Mid or Rid is blank / enter some text in work information / Save / see message "The member "X" was changed successfully."

### DIFF
--- a/src/member/models.py
+++ b/src/member/models.py
@@ -96,10 +96,12 @@ class Member(models.Model):
         verbose_name_plural = _('members')
 
     mid = models.IntegerField(
+        blank=True,
         null=True
     )
 
     rid = models.IntegerField(
+        blank=True,
         null=True
     )
 


erozqba:
Hi @lamontfr 
I think you are missing a migration file, you have modified the member model, so you should have a migration field generated, not?

About the migrations, there are another PR #792 that already add some missing migrations, I suggest to review and merge #792 first.

lamontfr:
@erozqba You are right. I am closing this PR and will make a new one so that the migrations are sequential.

PR #802, opened 2017-08-16
In model member, allow blanks in fields mid and rid

### LINKED ISSUE #796: Django Admin must accept a member with blank Mid and Rid fields

lamontfr:
- src/member/models.py : allow blank values
- src/member/migrations/0034_auto_20170816_0850.py

### Status

- [X] READY

### How to verify this change
- Admin / Members / select a member X / check that Mid or Rid is blank / enter some text in work information / Save / see message "The member "X" was changed successfully."


erozqba:
:+1:

PR #804, opened 2017-08-18
Add field relationships to member.Client

### LINKED ISSUE #793: Export client list to CSV gives error message
### Expected Behaviour
- When we display all clients as a list `/member/list/?status=A&display=list` the export button should produce a csv file of all the filtered to download.

### Actual Behaviour
- When we display all clients as a list `/member/list/?status=A&display=list` the export button gives the error "AttributeError at /member/list/ 'Client' object has no attribute 'relationships' "
- This may be related to issue #791

lamontfr:
- src/member/models.py : Client model
- src/member/migrations/0035_client_relationships.py : client table

### Status
- [X] READY

### How to verify this change
- Clients / click list display icon at top right / scroll to bottom of page / Export / see downloaded .CSV file.


erozqba:
Hi @lamontfr 
I have tested your changes, but I think that a different solution would be more appropriated. There is already a relationship between the Client and the Relationship models, so we don't really need to add another one, we should only use the existent, please look at the PR #806.

lamontfr:
@erozqba Thanks for your review and your better solution. I merged your PR #806.

PR #823, opened 2017-12-11
Command to set orders status to Delivered

### LINKED ISSUE #822: Automatically set status to "Delivered" for yesterday's orders
### Expected Behaviour
- The Billing only takes into account orders that have status equal to "Delivered".
- The Dashboard shows the number of meals delivered so far this month based on the order status.
- After midnight but before 8:00 AM, the Customer Manager wants the status to become "Delivered" for all orders for the preceding day (yesterday) that have the status "Ordered". The orders of yesterday that have any other status than "Ordered" will remain unchanged, for example the "No Charge" or the "Cancelled" orders.
- Before running the Billing process, the user can manually change the status of some orders to exclude them from the Billing.
### Actual Behaviour
- I have to manually change the status to become "delivered" for each single order that was actually delivered to customers; this is a very tedious process.
### Proposed solution
- Add a script to the nightly cron jobs.

lamontfr:
* src/order/management/commands/setordersdelivered.py : new command
* src/order/models.py : new constant
* src/order/tests.py : new unit test

### Status
- [X] READY

### How to verify this change
- In Souschef, Create a few orders for a date YYYY-MM-DD
- In Souschef, Orders / Reset / View (eye icon) / Pull down arrow at top right : Change the status of one of these orders for date YYYY-MM-DD to something different from "Ordered"
- In Souschef, Orders / Reset and notice which orders have status "Ordered" for date YYYY-MM-DD
- In the docker bash shell, run : ``` $ python3 manage.py setordersdelivered YYYY-MM-DD```
- In Souschef, Orders / Reset and notice that the status of orders that had status "Ordered" for date YYYY-MM-DD is now "Delivered".


src/order/management/commands/setordersdelivered.py
```diff
@@ -0,0 +1,38 @@
...
+        delivery_date = datetime.strptime(
+            options['delivery_date'], '%Y-%m-%d'
+        ).date()
+
+        orders = Order.objects.filter(
+            status=ORDER_STATUS_ORDERED,
+            delivery_date=delivery_date)
+        for order in orders:
+            order.status = ORDER_STATUS_DELIVERED
+            order.save()
```
erozqba: Why not use the update function directly? https://docs.djangoproject.com/en/dev/ref/models/querysets/#update

```
orders = Order.objects.filter(
    status=ORDER_STATUS_ORDERED,
    delivery_date=delivery_date
).update(
    status = ORDER_STATUS_DELIVERED
)
```

lamontfr:
@erozqba Thanks for your review and advice. I had initially thought of using an update but was not sure it was proper style.

Issue #842, opened 2019-01-08
`get_client_prices`


kousu:
While reviewing #839 I noticed `get_client_prices()`. It is a little shim to handle differently-billed clients but whereas it's called in `auto_create_orders()` and `create_batch_orders()`, in `add_item()`, which I believe is function called when entering orders manually, `MAIN_PRICE_DEFAULT` is used directly

https://github.com/savoirfairelinux/sous-chef/blob/396f86d696fd1dc541b6556dacea8a621e2b7420/src/order/models.py#L402

so I theorize it won't properly bill low-income clients if they call in an order.

https://github.com/savoirfairelinux/sous-chef/blob/396f86d696fd1dc541b6556dacea8a621e2b7420/src/order/models.py#L136-L146

Related issues:

* I think "SOLIDARY" is a typo; it should be "SOLIDARITY", as in, "prix solidaire" or "prix de solidarité"
* Apparently there are no clients using solidarity pricing at this time, so that part is dead code. Maybe it would be better just to remove it?
* `auto_create_orders()` and `create_batch_orders()` might be duplicates; I haven't traced the codebase yet to find out, but they sure sound like they are doing the same; I know that `auto_create_orders` is what `manage.py generateorders` ends up running, so does that mean that `create_batch_orders()` is dead code?

JeanFrancoisFournier:
Not sure but create_batch_orders() could be related to the ability within the UI to select multiple dates when creating an order manually. This would allow the user to speed up the order taking by applying the same items and prices to multiples individual orders. Also it might speed up the computing instead of running the price check for every individual order when doing this in batch? Not sure...

Related issues:
Solidary rate was not a typo, it's the way they used to call that higher braket in the 2015 sliding scale introduction. It's not used at this point but could allow us to add a 3rd level to our existing sliding scale when we get there.

Does this help?

PR #845, opened 2019-04-04
Flush unused methods and tests

manumilou:
## Tackles #842 

It turns out that `add_item` is not used, so the clients are always billed with the proper prices. It must have been part of a refactoring initiative that did not finish. 

🔥 

* Remove `add_item`, `remove_item` and `cancel` methods
* Remove corresponding test cases. `remove_item` and `cancel` were not tested.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### Additional notes

I let you decide if you want to merge that or not. In my opinion, it's better to remove dead code, as it can lead to confusion.

### DIFF
--- a/src/delivery/tests.py
+++ b/src/delivery/tests.py
@@ -156,58 +156,6 @@ class KitchenCountReportTestCase(SousChefTestMixin, TestCase):
         response = self.client.get('/delivery/viewDownloadKitchenCount/')
         self.assertTrue('ReportLab' in repr(response.content))
 
-    def test_extra_similar_side_dishes(self):
-        """Test cumulative quantities for similar side dishes."""
-        # generate orders today
-        self.today = datetime.date.today()
-        clients = Client.active.all()
-        numorders = Order.objects.auto_create_orders(
-            self.today, clients)
-        Menu.create_menu_and_components(
-            self.today,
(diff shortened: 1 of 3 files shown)


src/delivery/tests.py
```diff
@@ -156,58 +156,6 @@ def test_pdf_report_show_restrictions(self):
...
-            component_group='compote',
-            total_quantity=1,
-            remark='no sugar')
-        order.add_item(
-            'meal_component',
-            component_group='compote',
-            total_quantity=1,
-            remark='no sugar')
-        response = self.client.get('/delivery/kitchen_count/')
-        self.assertTrue(b'Compote' in response.content)
```
manumilou, line 209: I removed the entire test, as this is the only assertion. That's a lot of setup for a single test.

src/order/tests.py
```diff
@@ -39,35 +39,6 @@ def setUpTestData(cls):
         """
         cls.order = OrderFactory(order_item=None)
 
-    def test_order_add_item_meal_component(self):
```
manumilou, line 42: I did not remove the entire test class, as it is reused later in this test file. But, effectively, there is no more test in this class.

